PR #47, opened 2024-02-14
Bug 1880330 - only link to dashboards for supported surfaces

dmose:
Previously, we were linking to dashboards that didn't support the surfaces of the messages (i.e., linking to incorrect non-result dashboards).  This removes those links.


AllegroFox (approved):
LGTM!

PR #48, opened 2024-02-14
Bug 1880335 - used proposedDuration when needed

dmose:
Many (most? all?) live experiments have their endDate unset in the JSON.  However, we do want to display the proposed end date for live experiments. This computes that date using the startDate and the proposedDuration, and causes that to be rendered in the dashboard if the endDate is not present.

### DIFF
--- a/__tests__/lib/experimentUtils.test.ts
+++ b/__tests__/lib/experimentUtils.test.ts
@@ -1,0 +1,48 @@
+import { getProposedEndDate } from '../../lib/experimentUtils.ts'
+
+describe('getProposedEndDate', () => {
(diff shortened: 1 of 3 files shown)


AllegroFox (approved):
Looks great :)

AllegroFox:
As a future enhancement, maybe we could split the Date column into Start Date & End Date or otherwise distinguish them visually (that is unless we're planning to move toward the idea of the progress bar/circle icon we had in some of the proposals).

dmose:
Thanks!  I already have a patch in progress to use a basic version of the date styling shown in the figma.

PR #51, opened 2024-02-15
Bug 1880564 - make Skylight exp dates skimmable

dmose:
This makes the dates more readable and skimmable.  It also contains some type fixes to match the NimbusExperiment types used for dates.


AllegroFox (approved):
Nice!

PR #59, opened 2024-02-21
Bug 1880334 - Add infobar dashboards

dmose:
This cleans up the dashboard link creation code, and adds a link to the newly created infobar dashboard (based on the messaging explore, unlike the about:welcome dashboards, which are based on the event counts explore.

It would be great to have once over on the dashboard itself as part of this review to convince ourselves that it's showing accurate stuff.

### DIFF
--- a/lib/messageUtils.ts
+++ b/lib/messageUtils.ts
@@ -28,10 +28,25 @@ export function getTemplateFromMessage(msg : any) : string {
   return msg.template;
 }
 
-export function isAboutWelcomeTemplate( template : string ) : boolean {
+export function _isAboutWelcomeTemplate( template : string ) : boolean {
   // XXX multi shouldn't really be here, but for now, we're going to assume
   // it's a spotlight
   const aboutWelcomeSurfaces = ['feature_callout', 'multi', 'spotlight']
 
   return aboutWelcomeSurfaces.includes(template);
+}
+
+export function getDashboard( template: string, msgId: string ) : string | undefined {
+  const encodedMsgId = encodeURIComponent(msgId);
+  const encodedTemplate = encodeURIComponent(template);
+
+  if (_isAboutWelcomeTemplate(template)) {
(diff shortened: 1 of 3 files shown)


dmose:
One thing I should mention, is that unlike the other dashboard (which counts users), this one counts messages.  I suspect we'll want to talk to product/data to understand which one will be most useful...

CC @rfambro

dmose:
(Though I don't think we need to block landing this on having that discussion.

lib/messageUtils.ts
```diff
@@ -28,10 +28,25 @@ export function getTemplateFromMessage(msg : any) : string {
...
+  const encodedTemplate = encodeURIComponent(template);
+
+  if (_isAboutWelcomeTemplate(template)) {
+    return `https://mozilla.cloud.looker.com/dashboards/1471?Message+ID=%25${encodedMsgId?.toUpperCase()}%25`
+  }
+
+  if (template === 'infobar') {
+    return `https://mozilla.cloud.looker.com/dashboards/1622?Messaging+System+Ping+Type=${encodedTemplate}&Submission+Date=30+days&Messaging+System+Message+Id=${encodedMsgId}&Normalized+Channel=release&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=`
+  }
+
```
AllegroFox, line 50: This works! I feel like returning the URLs like this is likely to get brittle at some point though given how Looker works I don't know if there's a better way. I guess that since we're still discussing exactly what kind of dashboards we want, we'll be revisiting this part anyway.

dmose: That's an interesting point. I'd also be interested to see if this actually happens (eg most of the params seem unlikely to change).  That said, I could imagine the dashboard number maybe changing, though I'm not sure when/if that would happen.  Did I describe the mode(s) of brittleness you had in mind, or were you thinking of something else?

AllegroFox (approved):

PR #76, opened 2024-03-01
Show preview button for spotlight/infobar in experiments

AllegroFox:
Enables the "Preview Link" cell for infobars and Spotlights in the experiments table. `encodeURIcomponent` should also help with some of the `btoa` issues we were having, and `_substituteLocalizations` comes from Limelight and should allow recipes that include localization objects to still be previewed. (`about:messagepreview` doesn't know about localization and will error out if it's included in the object.)


app/page.tsx
```diff
@@ -1,43 +1,99 @@
...
+  getProposedEndDate,
+  usesMessagingFeatures,
+} from "../lib/experimentUtils.ts";
 import Link from "next/link";
 
-function getASRouterLocalColumnFromJSON(messageDef: any) : FxMSMessageInfo {
-  let fxmsMsgInfo : FxMSMessageInfo = {
-    product: 'Desktop',
-    release: 'Fx 123',
+function _substituteLocalizations(values, localizations) {
```
dmose, line 24: Let's put this into experimentUtils.ts and add tests for at least important stuff.

app/columns.tsx
```diff
@@ -255,43 +256,44 @@ export const experimentColumns: ColumnDef<ExperimentAndBranchInfo>[] = [
...
-
-        :
-        <a
-          className={buttonVariants({ variant: "outline", size: "sm" })}
-          href={props.row.original.previewLink}
-          target="_blank">
-          Preview
-        </a> );
+    return (
+      copyPreviewLink ?
```
dmose, line 274: The other side of the fork (in the ASRouter table) is actually more up-to-date re-styling.  Can you move that code into a PreviewLinkButton component in its own file and then call it from both there and here?

app/page.tsx
```diff
@@ -18,8 +19,10 @@ function getASRouterLocalColumnFromJSON(messageDef: any) : FxMSMessageInfo {
     metrics: 'some metrics',
     ctrPercent: .5, // getMeFromLooker
     ctrPercentChange: 2, // getMeFromLooker
-    previewLink: `about:messagepreview?json=${btoa(JSON.stringify(messageDef))}`,
-  }
+    previewLink: `about:messagepreview?json=${encodeURIComponent(btoa(
```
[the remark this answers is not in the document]
dmose: Also, thanks for noticing this bug and fixing it here in the ASRouter table too!

app/page.tsx
```diff
@@ -145,8 +160,8 @@ function getExperimentAndBranchInfoFromRecipe(recipe: NimbusExperiment) : Experi
 
   let experimentAndBranchInfos : ExperimentAndBranchInfo[] = [];
   experimentAndBranchInfos =
-    ([experimentInfo] as ExperimentAndBranchInfo[])
-    .concat(branchInfos);
+  ([experimentInfo] as ExperimentAndBranchInfo[])
+  .concat(branchInfos);
```
dmose, line 164: Please indent these two last to lines, as they are hard to read as-is.

lib/experimentUtils.ts
```diff
@@ -65,4 +65,37 @@ export function getProposedEndDate (startDate : string | null, proposedDuration
...
\ No newline at end of file
+}
+
+/**
+ * 
+ * @param values - the message contents
+ * 
+ * @param localizations - the localization object contained in the 
+ *                        experiment recipe. May be null.
+ */
```
dmose, line 76: Please add a detailed description of what the function is supposed to do.

Since we're adding new code here, can you write at least a few tests for the most important cases?

lib/experimentUtils.ts
```diff
@@ -65,4 +65,37 @@ export function getProposedEndDate (startDate : string | null, proposedDuration
...
+    return values;
+  }
+
+  if (Array.isArray(values)) {
+    return values.map((value) => _substituteLocalizations(value, localizations));
+  }
+
+  const substituted = Object.assign({}, values);
+
+  for (const [key, value] of Object.entries(values)) {
```
dmose, line 92: A comment above this for loop would be helpful for readers.

dmose (changes requested):
This looks great -- thanks for finding and fixing the encodeURIComponent bug in both places!  Just a few requests before landing-ready...

dmose:
Also, check https://github.com/mozilla/skylight/actions/runs/8116013167/job/22185107242?pr=76 from the CI.  For some reason (and I see this too), certain type-checking errors get caught by Netlify's CI, and by the eslint integration in VS Code, but not by the "npm run dev" and "npm run lint" steps.  I have yet to track this down.

dmose:
Looks like the errors are locally visible with `npm run build` but not with `npm run dev`.

dmose:
Yuck: intentional setup by next for performance: https://github.com/vercel/next.js/discussions/13617.   Easy enough to work around, though.

app/page.tsx
dmose: Also, adding types for this function would be worthwhile, and might solve some of the complaints from the compiler in `npm run build`.

app/columns.tsx
AllegroFox: Can do!

app/page.tsx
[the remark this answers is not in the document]
AllegroFox: done!

app/page.tsx
```diff
@@ -57,7 +59,8 @@ function getBranchInfosFromExperiment(recipe: NimbusExperiment) : BranchInfo[] {
       case 'infobar':
         branchInfo.id = value.messages[0].id
         branchInfo.ctrDashboardLink = getDashboard(template, branchInfo.id)
-        break
+        branchInfo.previewLink = getPreviewLink(_substituteLocalizations(value.content, recipe.localizations?.[Object.keys(recipe.localizations)[0]]));
```
AllegroFox, line 62: This `[Object.keys(recipe.localizations)[0]]` thing is kinda gross but it means we correctly access the first locale inside the localization object. We'll probably want to add a dropdown component that allows us to choose a locale from the available ones, to pass to this function.

dmose: How about adding this in as an XXX comment?

AllegroFox: Will do :)

dmose: Also, please try to avoid requiring readers to make mental context switches inside of a single line of code.  I.e. split out the `_substituteLocalizations` function call to at least one other line and a temporary variable, maybe more, depending on how readable it feels.

dmose: Another way to win on readability here might be to add a 'getFirstLocalization' method in experimentUtils.ts.  Your call.

app/page.tsx
```diff
@@ -67,21 +70,25 @@ function getBranchInfosFromExperiment(recipe: NimbusExperiment) : BranchInfo[] {
...
+	    const firstMessage = value.messages[0]
         if (!('content' in firstMessage)) {
-	  console.warn('template "multi" first message does not contain content key; details not rendered')
-	  return branchInfo
+      	  console.warn('template "multi" first message does not contain content key; details not rendered')
+      	  return branchInfo
         }
 
-	// XXX only does first screen
+	      // XXX only does first screen
```
dmose, line 86: Outdent to match the context, please.

app/page.tsx
```diff
@@ -114,12 +121,13 @@ function getBranchInfosFromExperiment(recipe: NimbusExperiment) : BranchInfo[] {
...
+function getExperimentAndBranchInfoFromRecipe(recipe: NimbusExperiment): ExperimentAndBranchInfo[] {
 
   // console.log("in gECFJ");
-  if (recipe.isRollout) {
-    return [];
-  };
+  // there are no non-rollout spotlights right now, we can comment this out to test them
+  // if (recipe.isRollout) {
+  //   return [];
+  // };
```
dmose, line 130: If you could revert the commenting here, that'd be great.

lib/experimentUtils.ts
```diff
@@ -65,4 +65,50 @@ export function getProposedEndDate (startDate : string | null, proposedDuration
...
+ * Right now, we are manually passing the first locale out of the localization object.
+ * Eventually we'll want to select a locale on the dashboard (probably with a dropdown.)
+ *
+ * @param values - The values to perform substitutions upon; message contents
+ * @param localizations - The localization object from the recipe, contains
+ *        substitutions for a specific locale. May be null.
+ * @returns {any} The values, potentially locale substituted.
+ */
+
+export function _substituteLocalizations(values: any, localizations: any) : object {
```
dmose, line 89: How about copying [the localizations type from nimbus shared](https://github.com/mozilla/nimbus-shared/blob/main/types/experiments.ts#L121-L131)?

I'm not quite sure what to do about the values typing. If there's something we could copy-paste, that'd be great.  If not, c'est la vie. Up to you...

lib/messageUtils.ts
```diff
@@ -50,3 +50,11 @@ export function getDashboard( template: string, msgId: string ) : string | undef
 
   return undefined
 }
+
+export function getPreviewLink( message: any ) : string | undefined {
```
dmose, line 54: It seems like this function would never actually return undefined, would it?

If not, can we remove it from the type signature?

dmose (approved):
Looks great; thanks for the nice work!  Just some small bits to tidy up before landing...

r=dmose once you've made those tweaks

app/page.tsx
AllegroFox: Looks like there was a weird combo of tabs and spaces here. Fixed!

lib/messageUtils.ts
AllegroFox: yep!

PR #77, opened 2024-03-03
Now that Netlify supports Node 20, switch to it

dmose:
I had to waste a non-trivial amount of time fighting with node versioning in a different project, forcing me to clean up my local setup. Once that was done, it was trivial to update this project to be less error-prone by making the versioning stricter and automatically configuring it to use the correct version for developers who use [`asdf`](https://asdf-vm.com/).


AllegroFox (approved):
Thanks for doing this! WFM

PR #82, opened 2024-03-05
Bug 1883576 - Show JSON/screenshots for branches w/o previews

dmose:
This is rebased on a copy of the infobar-preview branch, itself rebased against the latest main, which is why you'll see all the commits from infobar-preview at the head of this PR.


app/columns.tsx
```diff
@@ -218,19 +220,11 @@ export const experimentColumns: ColumnDef<ExperimentAndBranchInfo>[] = [
...
-  }, {
-    accessorKey: "previewLink",
-    header: "",
-    cell: (props: any) => {
       if (props.row.original.previewLink == undefined) {
-        return ( <div/> );
+        const branchLink = getBranchLink(props.row.original.recipe, props.row.original.slug);
+        return (
+          OffsiteLink(branchLink, "JSON/Screenshots")
+        )
```
AllegroFox, line 227: How do we want to combine this with the actual preview link? I'd say it's equally valid to do two different columns (We may want to link to the JSON even if a preview is available) or to only show the screenshots link if the preview isn't present.

dmose: Those are good questions that I hadn't really thought about.  How about we wait to see what feedback from users looks like (e.g., if they say they miss one when it's not there or ask to see both)?

dmose: (There is, after all, a complexity tradeoff to showing more affordances on the screen at once).

dmose:
The experimenter changes I landed in mozilla/experimenter#10368 have been deployed, so I've pushed a commit to this PR which links directly to branch screenshots rather than the branch summary.  The UX for when there aren't any screenshots (eg the control branch of many experiments) is confusing (it just links to the top of the experiment summary page), but the overall experience seems better on balance.  We'll need an API of some kind to fix this; I'll file an experimenter ticket for that.

dmose:
To get a sense for the updated UX, click on the "Deploy Preview" near the top of this PR.

AllegroFox (approved):
Works for me!

PR #95, opened 2024-03-11
Bug 1883820 - refactor experiment code into NimbusRecipe for clarity, part 1

dmose:
[Bug 1883820](https://bugzilla.mozilla.org/show_bug.cgi?id=1883820)

This should help make the overall code substantially more readable and writable.

This code should be easier to review than it might appear at first glance:

`page.tsx:getBranchInfosFromExperiment` -> `NimbusRecipe.getBranchInfos`
is entirely about moving a function from one file to a method in another file.
There are almost no changes here, other than whitespace, and use of `this._rawRecipe`.

`page.tsx:getExperimentAndBranchInfoFromRecipe` -> 
`NimbusRecipe.getRecipeInfo`, `NimbusRecipe.getRecipeOrBranchInfos`

Is extremely similar, although it extracts the first piece of code to its own method for clarity.

There is a second PR in progress to split NimbusRecipe.getBranchInfos apart for similar clarity, but that can be reviewed and landed separately.


AllegroFox (approved):
WFM!

PR #98, opened 2024-03-12
Bug 1883820 - refactor experiment code into NimbusRecipe for clarity, part 2

dmose:
More cleanup.  I'm not actually expecting to merge this into the `nimbus-recipe` branch, but instead into `main` after the `nimbus-recipe` branch gets merged in.  This is a separate PR so that the review can be chunked up into smaller pieces.


AllegroFox (approved):
WFM, thanks!

PR #115, opened 2024-03-21
Support about:welcome experiments and preview

AllegroFox:
Not a spike any more! Previews about:welcome screens by shoehorning them into a Spotlight-shaped wrapper. Also fixes an issue with localized messages where previews would break on locales with non-latin text (ru, zh-CN) by adding some binary encoding to the preview links.


lib/messageUtils.ts
```diff
@@ -1,60 +1,78 @@
...
-export function _isAboutWelcomeTemplate( template : string ) : boolean {
+export function _isAboutWelcomeTemplate(template: string): boolean {
   // XXX multi shouldn't really be here, but for now, we're going to assume
   // it's a spotlight
-  const aboutWelcomeSurfaces = ['feature_callout', 'multi', 'spotlight']
+  const aboutWelcomeSurfaces = [
+    "feature_callout",
+    "multi",
+    "spotlight",
+    "aboutwelcome",
```
dmose, line 38: Move this to the top of the list as well, please.

lib/nimbusRecipe.ts
```diff
@@ -41,12 +42,40 @@ export class NimbusRecipe implements NimbusRecipeType {
...
     // XXX in this case we're really passing a feature value. Hmm....
-    const template = getTemplateFromMessage(featureValue)
-    branch.template = template
-    branchInfo.template = template
-    branchInfo.surface = getDisplayNameForTemplate(template)
+    // XXX about:welcome is special, so we have to account for it
+    let template;
+    (branch.features[0].featureId === "aboutwelcome" && branch.slug != 'control')
+      ? (template = "aboutwelcome")
+      : (template = getTemplateFromMessage(featureValue));
```
dmose, line 49: If we can do this as `if/then/else` blocks, it would improve readbility.

dmose: Also, why is `control` special cased here?  If it really should stay that way, it would be good to have a comment explaining the reasoning.

lib/nimbusRecipe.ts
```diff
@@ -41,12 +42,40 @@ export class NimbusRecipe implements NimbusRecipeType {
...
+        spotlightFake.content.backdrop =
+          "var(--mr-welcome-background-color) var(--mr-welcome-background-gradient)";
+        // Localize the recipe if necessary.
+        let localizedWelcome = _substituteLocalizations(
+          spotlightFake,
+          this._rawRecipe.localizations?.[
+            Object.keys(this._rawRecipe.localizations)[0]
+          ],
+        );
+        // XXX assumes previewable message (spotlight?)
```
dmose, line 75: Can this comment line be deleted?

lib/nimbusRecipe.ts
```diff
@@ -41,12 +42,40 @@ export class NimbusRecipe implements NimbusRecipeType {
...
+    (branch.features[0].featureId === "aboutwelcome" && branch.slug != 'control')
+      ? (template = "aboutwelcome")
+      : (template = getTemplateFromMessage(featureValue));
+
+    branch.template = template;
+    branchInfo.template = template;
+    branchInfo.surface = getDisplayNameForTemplate(template);
+
+    switch (template) {
+      case "aboutwelcome":
```
dmose, line 56: Some tests for this stanza would be good, particularly around things that seem unlikely to change much.

dmose (changes requested):
In general, this looks great! A decent chunk of this is prettier-based reformatting, which I don't think we should try to fix for this patch because it doesn't seem worth the effort. 

1. Can you turn off the prettier formatting in your editor for just this repo until we have a shared prettier config?
2. Can you please file a bug/ticket about the experiment dashboard links needing to restrict the results to the specific experiment given common branch message IDs like `treatment-a` and `treatment-b`?
3. I don't think we necessarily need to fix that bug before we land this, but we need to do something. Maybe we should change the dashboard links to a non-clickable "Dashboard coming soon" string for the' about: welcome' experiments (or all experiments?).

lib/nimbusRecipe.ts
AllegroFox: Since we're just directly assigning the template instead of doing `getTemplateFromMessage`, the control branch ends up with a template instead of "none" in the table. (We can't use the function for about:welcome because it doesn't actually use the `template` property on its recipes. Go figure)

lib/nimbusRecipe.ts
AllegroFox: yup!

lib/messageUtils.ts
```diff
@@ -1,60 +1,78 @@
...
+  if (template === "infobar") {
+    return `https://mozilla.cloud.looker.com/dashboards/1622?Messaging+System+Ping+Type=${encodedTemplate}&Submission+Date=30+days&Messaging+System+Message+Id=${encodedMsgId}&Normalized+Channel=release&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=`;
   }
 
-  return undefined
+  return undefined;
+}
+
+// convert a UTF-8 string to a string in which each 16-bit unit occupies
+// only one byte. This is necessary for non-latin characters.
```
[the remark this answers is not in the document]
dmose: I think my original comment is likely to have been mistaken; I'll dig into this more on Monday.

AllegroFox:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1887077 && set the experiment dashboards to an unclickable "Dashboard coming soon" in the meantime.

lib/messageUtils.ts
```diff
@@ -1,60 +1,79 @@
...
```
dmose, line 64:
```suggestion
// Convert a UTF-8 string to a string in which only one byte of each 
// 16-bit unit is occupied. This is necessary to comply with `btoa` API constraints.
```

lib/messageUtils.ts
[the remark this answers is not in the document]
dmose: My mistake; this should be fine.  I've proposed a clarification of the comment.

dmose (approved):
This looks great; thanks for your hard work!  r=dmose; please address the few nits that came up before merging.

__tests__/lib/nimbusRecipe.test.ts
```diff
@@ -61,6 +110,28 @@ describe('NimbusRecipe', () => {
...
+    it('returns a specialized BranchInfo object if the recipe is from about:welcome', () => {
+      const nimbusRecipe = new NimbusRecipe(AW_RECIPE)
+
+      const branch = AW_RECIPE.branches[1]
+
+      const branchInfo = nimbusRecipe.getBranchInfo(branch)
+
+      // XXX getBranchInfo is actually going to return a previewLink, which
+      // makes this test kind of brittle. We could refactor this to no longer 
+      // use deepEqual and check for the existence of object properties instead.
```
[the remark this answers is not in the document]
AllegroFox: yeah that's basically it. Any change to the shape of NimbusRecipe or getBranchInfo will break the test, since it's testing for the exact encoded previewLink. It's probably not that big of a deal as long as we know about it, but we could refactor if it starts to get annoying.

lib/nimbusRecipe.ts
```diff
@@ -41,12 +42,44 @@ export class NimbusRecipe implements NimbusRecipeType {
...
 
     // XXX in this case we're really passing a feature value. Hmm....
-    const template = getTemplateFromMessage(featureValue)
-    branch.template = template
-    branchInfo.template = template
-    branchInfo.surface = getDisplayNameForTemplate(template)
+    // about:welcome is special and doesn't use the template property,
+    // so we have to assign it directly to treatment branches
+    let template;
+    if (branch.features[0].featureId === "aboutwelcome" && branch.slug != 'control') {
```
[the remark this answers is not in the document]
AllegroFox: 👍

lib/messageUtils.ts
AllegroFox: done!

__tests__/lib/nimbusRecipe.test.ts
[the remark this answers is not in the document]
dmose: Yeah, let's see if it turns out to be worth tweaking.

lib/nimbusRecipe.ts
```diff
@@ -59,27 +59,30 @@ export class NimbusRecipe implements NimbusRecipeType {
...
-        let localizedWelcome = _substituteLocalizations(
-          spotlightFake,
-          this._rawRecipe.localizations?.[
-            Object.keys(this._rawRecipe.localizations)[0]
-          ],
-        );
-
-        branchInfo.previewLink = getPreviewLink(localizedWelcome);
+        // Make sure there's a message to preview
+        if (featureValue.screens) {
```
dmose, line 63: Readbility nit: instead of indenting a big clause, how about:

```javascript
if (!featureValue.screens) {
  break;
}

dmose:
Looks good to me with one minor tweak; r=dmose.  Thanks again!

dmose (approved):

lib/nimbusRecipe.ts
AllegroFox: Agreed!

PR #127, opened 2024-03-28
Bug 1885962 - Display experiment count and multi-feature experiments

dmose:
This also adds a NimbusRecipeCollection and cleans up a bunch of code.

### DIFF
--- a/lib/experimentUtils.ts
+++ b/lib/experimentUtils.ts
@@ -34,17 +34,6 @@ export const MESSAGING_EXPERIMENTS_DEFAULT_FEATURES : string[] = [
   "spotlight",
 ];
 
-// XXX this should really be a method on NimbusRecipe
-export function usesMessagingFeatures(recipe : NimbusExperiment): boolean {
-  // XXX iterate through all features instead of just checking the first
-  // XXX figure out how to better handle the ?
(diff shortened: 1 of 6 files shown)


lib/nimbusRecipe.ts
```diff
@@ -60,20 +74,20 @@ export class NimbusRecipe implements NimbusRecipeType {
     switch (template) {
       case "aboutwelcome":
         // Make sure there's a message to preview, bail out early otherwise
-        if (!featureValue.screens) {
+        if (!feature.value.screens) {
           break;
         }
         // featureValue will become the "content" object in a spotlight JSON
```
AllegroFox, line 80: nit: change this comment to feature.value just to keep things consistent

lib/experimentUtils.ts
```diff
@@ -34,17 +34,6 @@ export const MESSAGING_EXPERIMENTS_DEFAULT_FEATURES : string[] = [
   "spotlight",
```
AllegroFox, line 34: Adding "testFeature" here will keep the tests happy, since we're filtering experiments now :)

AllegroFox (approved):
Looks good with one (teeny tiny) nit, R+ with the test fixed!

PR #130, opened 2024-03-29
Remove release column, clarify table headings, switch prod dashboards to release channel

dmose:
This also adds a tooltip explaining the Firefox version requirement.  

I've removed the release column infrastructure because there are too many things for "release" to mean; I don't think it will come back in the same UI form.

`getDashboard` now takes an optional `channel` argument; it is used to switch production dashboards to default to the release channel.


__tests__/lib/messageUtils.test.ts
```diff
@@ -56,3 +56,31 @@ describe('toBinary', () => {
     expect(decodedResult).toEqual(nonLatinString);
   })
 })
+
+describe('getDashboard', () => {
+  // Test that get dashboard returns a correct dashboard links for a given
```
AllegroFox, line 61: nit: comment typos :)

__tests__/lib/messageUtils.test.ts
```diff
@@ -56,3 +56,31 @@ describe('toBinary', () => {
...
+  // for a given messageId and template === "infobar".
+  it('returns a correct infobar dashboard link', () => {
+    const template = "infobar"
+    const msgId = "123"
+    const channel = "rel:ease"
```
AllegroFox, line 66: nit: lets add a note that extra characters are intentional to check the encoding

AllegroFox (approved):
Looks great, the changes to the table header make it much clearer :)

__tests__/lib/messageUtils.test.ts
dmose: I actually just deleted this and the other comment, but they are both just repeating what is already in the `it()` clause directly below them.

PR #133, opened 2024-04-02
Bug 1889264 - Add a CHANGELOG.md

dmose:
Here's a basic changelog for the project.  It doesn't capture every little thing (eg dependency updates, etc), and we'll have to decide what we want to do about this moving forward, but it feels like a good start.

CC @AllegroFox @rfambro

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,69 @@
+## Monday, April 1st, 2024
+
+### Added
+
+* Display about:welcome experiments (we now show all messaging experiments!)
+* Add Previews for about:welcome experiments.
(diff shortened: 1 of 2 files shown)


dmose:
I'm inclined to add a skeleton CONTRIBUTING.md file which (to start) says that almost all PRs should add a CHANGELOG.md entry.  Thoughts?

AllegroFox:
> I'm inclined to add a skeleton CONTRIBUTING.md file which (to start) says that almost all PRs should add a CHANGELOG.md entry. Thoughts?

I'll need that reminder myself so I am in favour of this :)

AllegroFox (approved):
Thanks!

PR #135, opened 2024-04-02
Bug 1889195 - Fix empty dashboards caused by upcasing msgId

dmose:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1889195 for details

### DIFF
--- a/__tests__/lib/messageUtils.test.ts
+++ b/__tests__/lib/messageUtils.test.ts
@@ -73,7 +73,7 @@ describe('getDashboard', () => {
 
   it('returns a correct featureCallout dashboard link', () => {
     const template = "feature_callout"
-    const msgId = "1:23" // weird chars to test URI encoding
+    const msgId = "test1:23" // weird chars to test URI encoding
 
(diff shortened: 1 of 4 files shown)


dmose:
I'm changing this to a full-on PR, ready for review.  Either before that review, or as part of it, we need to do the Bugzilla / github / searchfox archeology mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1889195#c0 to be sure that we're not going to break anything here.

dmose:
Move this back to draft - it fixes some dashboards while breaking others.  Needs more investigation.

dmose:
As an example, the dashboard for https://experimenter.services.mozilla.com/nimbus/np-survey-new-vs-existing-vs-returning-ru-zh-cn breaks.

I've figured out how to create a custom dimension of the lowercased message id, because I was hoping to do a case-insensitive compare on the Looker side.  Unfortunately, such custom dimensions apparently can't be used as filters at the dashboard level itself.

One could imagine making the message-id field case-insensitive at the explore level, that probably requires more QA and discussions with Looker folks, and might be the right way to go.

Perhaps there's another workaround, however.

dmose:
I've just tested on `main`, and the dashboard mentioned in https://github.com/mozilla/skylight/pull/135#issuecomment-2045587369 has already been fixed.  The other dashboards mentioned in the bug and currently not being shown will be reenabled by #432.

PR #157, opened 2024-04-11
Bug 1891093 - Switch from Remote Settings to Experimenter API

dmose:
Bug 1891093 - switch from pulling info from remote settings to pulling it from the Experiment API so that we can get user-friendly branch names and fix broken screenshot links.

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,10 @@
 # .env
 
-# Remote settings records URL
-RECORDS_URL="https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/"
+# Base URL for the experimenter API (defaults to the production instance)
+#
+EXPERIMENTER_API_PREFIX="https://experimenter.services.mozilla.com/api/v7/"
 
-#Collection 
-COLLECTION="nimbus-desktop-experiments/records"
+# API call with parameters to fetch experiments we want to display.
+# https://htmlpreview.github.io/?https://github.com/mozilla/experimenter/blob/main/docs/experimenter/swagger-ui.html has more info.
+#
+EXPERIMENTER_API_CALL="experiments/?status=Live&application=firefox-desktop"
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,4 @@
 # Copy this to .env.local to set local variables
 
-# Nimbus recipe collection (defaults to prod)
-
 # Preview
-# COLLECTION="nimbus-preview/records"
+# EXPERIMENTER_API_CALL="experiments/?status=Preview&application=firefox-desktop"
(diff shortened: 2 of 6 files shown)


dmose:
I remember I had an old branch where I played around with trying to use the Experimenter API (which is where I discovered the cache issue I mentioned in the Bugzilla bug.

I brought that patch forward and it seems to basically work.

Still needs to (at least) fix existing tests, probably add at least one new test, and figure out why the number of experiments between the code on `main` and the code on this branch is different.

dmose:
I don't think we need an additional test, since we're already testing the unit where we call `fetch()`.  The CI here won't pass and this won't be testable until mozilla/experimenter/pull/10648 lands, so we may need to retrigger here after it does.

AllegroFox:
I'm seeing an error in the JSON response: 
`Error: Unexpected token '<', "<!--
     "... is not valid JSON`

Could it be that the updated experimenter hasn't deployed yet, or is there a config issue?

dmose:
It turns out that https://github.com/mozilla-services/cloudops-infra/pull/5610 also has to merge.  The thread in nimbus-dev that I started on Friday has more details.

dmose:
Although that PR has been merged and deployed to production, the v7 API isn't yet public for some reason.  I've pinged in,https://mozilla-hub.atlassian.net/browse/DSRE-1622...

dmose:
OK, the CI cache is cleared, and this baby is ready for review.

dmose:
CC @sarahhjchung

aminomancer (approved):
Works for me and code looks good.

I noticed screenshots are included in the branch values, so is there any expectation that we will need to use private API methods? I'm surprised all this can be accessed with the basic public method, without SSO or API keys, since the rv includes a lot of information that is normally locked behind SSO in the Experimenter UI (recipe pages are entirely inaccessible). Do we know whether the Nimbus team is okay with us making all this info effectively public?

dmose:
> I noticed screenshots are included in the branch values, so is there any expectation that we will need to use private API methods? 

Not at this moment, but I won't be surprised if that changes in the future, depending on the information we ask them to expose.

> since the rv includes a lot of information that is normally locked behind SSO in the Experimenter UI (recipe pages are entirely inaccessible)

The return value should (currently) be effectively the same (or maybe a subset) of the info exposed by the v6 API, which is really just the same as what already gets publicly distributed to remote settings... with the exception of also including screenshot URLs and branch descriptions, which I wouldn't expect to be particularly private.

> Do we know whether the Nimbus team is okay with us making all this info effectively public?

Yes, Jared (the team lead) both wrote the code and filed the ops ticket to make it public.

aminomancer:
Awesome, that all makes sense.

PR #178, opened 2024-04-24
Bug 1888930 - Create env file && documentation for local variables

AllegroFox:


.env.sample
```diff
@@ -0,0 +1,15 @@
...
+# Select an experiment collection
+# Prod link
+RECORDS_URL="https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/nimbus-desktop-experiments/records"
+
+# Preview link
+# RECORDS_URL="https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/nimbus-preview/records"
+
+# Secrets
+EXPERIMENTER_API_KEY=
+NODE_ENV="development"
```
dmose, line 12: Let's wait on adding the EXPERIMENTER_API_KEY, as I'm not sure that's exactly how OAuth Connect works, and we don't need it yet anyway.

Maybe change "Secrets" to "Misc" for now?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Wednesday, April 24th, 2024
+
+### Added
+
+* support for a local .env file
```
dmose, line 5: Make this a bit more explicit?  (e.g., "get configurables from the environment, and provide a sample .env file which can be copied for local development").

README.md
```diff
@@ -7,6 +7,17 @@ A bunch of the code, ideas, and links come from the OMC team's work week Hackath
 ```bash
 npm run dev
 ```
+## Setting local environment variables
```
dmose, line 10: This section needs to be before the "Running the development server" section, I think...

.env.sample
```diff
@@ -0,0 +1,15 @@
...
+
+# Dashboardy things
+FORCED_ENROLLMENT_PREFIX="optin-"
```
dmose, line 15: We don't appear to be using this.  Do we need it?

dmose (changes requested):
Be sure to also add the local stuff to .gitignore, a la https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables

.env.sample
AllegroFox: I added it thinking it would be the next thing I played around with. I'll keep it out of this patch for simplicity!

CHANGELOG.md
AllegroFox: Done!

AllegroFox:
> Be sure to also add the local stuff to .gitignore, a la https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables

`.env*.local` is actually already in the .gitignore, so we should be go to go there :)

.env.d.ts
```diff
@@ -0,0 +1,10 @@
+// Declare environment variable types
+
+namespace NodeJS {
+  interface ProcessEnv {
+    RECORDS_URL: string;
+    NODE_ENV: 'development' | 'production';
```
dmose, line 6: nextjs.org/docs/messages/non-standard-node-env makes it look to me like Next more or less manages NODE_ENV for us to some reasonable degree, so it may not make sense to encourage tweaking it by hand until we get to the point where think there's a good reason to do that. Was there some particular reason you included it? If not, maybe just remove it...

lib/nimbusRecipeCollection.ts
```diff
@@ -16,7 +16,7 @@ export class NimbusRecipeCollection implements NimbusRecipeCollectionType {
 
   async fetchRecipes() : Promise<Array<NimbusRecipe>> {
       const response = await fetch(
-      "https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/nimbus-desktop-experiments/records",
+      process.env.RECORDS_URL as string,
```
dmose, line 19: Since the type information is being effectively discarded by needing to use type assertion (and a reasonable code reader might assume that, given the presence of the type information, they're being protected by it), my gut is that it's not worth the carrying cost of leaving the info there now.

If we decide this is worth it at some point, we could use something like the `throwIfNot` approach from step 3 of [this stack overflow example](https://stackoverflow.com/a/56666712).

dmose (commented):
This is getting close to being ready to land, which is great.  I've added a few more comments/thoughts.  If you could also add the bug number to the PR, that'd be nice.

lib/nimbusRecipeCollection.ts
AllegroFox: I like this && I think I'll do it as a followup. Right now I'm forcibly stringifying the two variables together so the type checker is fine, but it'll be valuable to make sure the variables used are present at runtime.

dmose (approved):
Looks good; r=dmose

PR #196, opened 2024-05-03
Bug 1894966 - Removed topic column from skylight

sarahhjchung:
- Removed "topic" column from the skylight table
- Removed "topic" data from type definitions


dmose:
Looks good to me, and seems to be working nicely.  Thanks for the nice visual simplification and code cleanup! r=dmose

PR #200, opened 2024-05-06
Bug 1889271 - Make all multi-message/screen surfaces in Skylight include "(1st screen)"

sarahhjchung:
- Updated the `feature_callout` display name to include "(1st screen)" instead of "(screen 0)"
- Updated the `aboutwelcome` display name to include "(1st screen)"
- Updated unit tests that included the `aboutwelcome` surface name


dmose:
Looks good; thank you!

PR #204, opened 2024-05-08
Bug 1891100 - Only render existing screenshot links

sarahhjchung:
Bug 1891100

Changes made:
- Checks that the experiment branch's `screenshots` property is non-empty and only renders the screenshot links in that case

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -87,6 +87,7 @@ export type BranchInfo = {
   nimbusExperiment: NimbusExperiment
   isBranch?: boolean
   template?: string
+  screenshots?: string[]
 } | []
 
 export type RecipeOrBranchInfo = RecipeInfo | BranchInfo;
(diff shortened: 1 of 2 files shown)


app/columns.tsx
```diff
@@ -229,11 +230,11 @@ export const experimentColumns: ColumnDef<RecipeOrBranchInfo>[] = [
...
         const recipe = new NimbusRecipe(props.row.original.nimbusExperiment)
-        const branchLink = recipe.getBranchScreenshotsLink(
-            props.row.original.slug);
-        return (
-          OffsiteLink(branchLink, "Screenshots")
-        )
+        const branchLink =
+          props.row.original.screenshots.length > 0
+            ? recipe.getBranchScreenshotsLink(props.row.original.slug)
+            : null;
```
dmose, line 236: Please break this up into a few lines (maybe just use an `if` rather than the ternary operator, so that it's easier to read because one doesn't have to make so many mental context switches in a single statement.

dmose (changes requested):
This looks good; thanks for the patch! I've asked for one readability tweak. I'm a little tempted to ask for a unit test, but this doesn't seem like a likely to break area as a unit. 

Ultimately, I think we probably want some integration testing so that we could test whole dashboards at a time and would be more likely to catch issues, but I think that's out of scope for a bug of this size.  :-)

dmose (approved):
Looks good; r=dmose

PR #205, opened 2024-05-08
Implement Auth0 for Skylight

AllegroFox:
I thought we were good to go with this, but there was a suggestion to use a personal Auth0 account to test first before sending the integration ticket to Mozilla SSO, and having done that I'm seeing some issues.

The app builds and generates the appropriate routes, but I don't see the auth being used or validated (i.e. the dashboard renders anyway.)

I'm *thinking* that the issue is that we're using server components, and therefore I will probably have to implement some middleware. I'm working on that now, but if you spot something else please LMK! 

EDIT: No middleware! Just correct usage of `withPageAuthRequired` 😃


dmose (commented):
OK, I got it working!  As you say, it's finicky with the callbacks -- I tried to run a local copy on http://localhost:3001/ and that failed (even after a changing with a message

app/index.tsx
```diff
@@ -0,0 +1,21 @@
+'use client';
```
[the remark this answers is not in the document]
dmose: I could imagine putting this stuff in layout.tsx, so it would at least handle children of this route if we make any.

I suppose the safest thing to do might be to use middleware to require it for everything in the app.

I'm not sure any of these things need to be done at all, let alone now, I'm just trying to get my head around the options/tradeoffs here, so I'd love your thoughts.

dmose (commented):
Thanks for sorting through this stuff and making it go.  I've gotten it working locally, and have started playing around and poking through the SDK docs and other Next docs...

.env.sample
```diff
@@ -2,3 +2,19 @@
 
 # Preview
 # EXPERIMENTER_API_CALL="experiments/?status=Preview&application=firefox-desktop"
+
+# Auth0 properties
+# A long, secret value used to encrypt the session cookie
+AUTH0_SECRET='LONG_RANDOM_VALUE'
+
+# The base url of the application
+AUTH0_BASE_URL='http://localhost:3000'
```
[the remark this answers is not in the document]
AllegroFox: Yep, it has to match the callback URL hosted on Auth0, though it is easy to change!

app/index.tsx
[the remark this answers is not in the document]
AllegroFox: Yep when I finally did get it, I realized it wasn't routing through `Index()` at all. This whole file should be removed now!

app/page.tsx
```diff
@@ -76,7 +78,7 @@ async function getMsgExpRecipeCollection(): Promise<NimbusRecipeCollection> {
   return msgExpRecipeCollection
 }
 
-export default async function Dashboard() {
+export default withPageAuthRequired(async function Dashboard() {
```
dmose, line 81: I'm inclined to instead use the first example at https://github.com/auth0/nextjs-auth0/blob/main/EXAMPLES.md#protecting-pages-with-middleware to protect all the pages, so we can't inadvertently add an unprotected route by forgetting to wrap it.  What do you think?

AllegroFox: I actually did try this and was having server/client issues; I'll give it another shot :)

AllegroFox: I found the issue, I needed to do some extra `init` things! I believe it's working with the middleware now!

app/page.tsx
```diff
@@ -95,7 +94,7 @@ export default async function Dashboard() {
         <h4 className="scroll-m-20 text-3xl font-semibold text-center py-4">
           Skylight
         </h4>
-
+        <a className="scroll-m-20 mx-20 text-m" href="/api/auth/logout">Logout</a>
```
dmose, line 97: Please put this down at the bottom of the page somewhere, and in a smaller font.

dmose (changes requested):
This is looking better and better...

I think it would be good to create a (non-public) GDoc about our deployment with  documentation about how to set this up in production / who to contact / where to file tickets.

https://github.com/auth0/nextjs-auth0/blob/main/SECURITY.md#caching-and-security has me a bit concerned. My guess is that `withMiddlewareAuthRequired` means that we want to turn off caching too. Would you please look into this, make any necessary adjustments, and document them in the GDoc?

app/page.tsx
AllegroFox: Sure!

AllegroFox:
Having investigated a bit, I think I'd prefer to write tests in a separate PR, just to unblock this. LMK if that works :)

AllegroFox:
Looking into the caching too!

app/page.tsx
AllegroFox: I've thrown the button down at the bottom for now, we might want to do a design pass on that eventually. (The logout button is also not strictly necessary because the SSO should just expire - so maybe we could just drop it?)

dmose:
It might be that the simplest thing to do is just to use devtools and see if the Set-Cookie is identical as described in that doc...

AllegroFox:
> It might be that the simplest thing to do is just to use devtools and see if the Set-Cookie is identical as described in that doc...

I didn't even think of checking that way, I'll do that!

AllegroFox:
I've tested a few different ways and I see a different Set-Cookie header each time so I think we're OK! (Even between quick refreshes before the cookie would typically expire) I also see response headers marked `no-store, must-revalidate` (as they should be I think)

dmose:
Open questions:
* do we want deploys to require auth?  yes, if we don't have fake looker data for the CTR deploys
  * will that work?  we'll see...
* should it be possible to entirely disable auth0 for development? yes
* should it be disabled by default?
* link to (the private) deployment doc from somewhere in this tree?

AllegroFox:
I think (fingers crossed) we are in good shape now; the deploy previews are working again so the env variable seems to be having the desired effect :) Prod was tested yesterday with Van; we'll end-to-end test the integration again once the PR has landed. (And we can always turn this off again in code if need be.)

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ### Added
 
+* SSO integration using the nextjs-auth0 package
+* Middleware that conditionally loads the Auth0 wrapper based on an
```
dmose, line 6: I would probably put this second point under an `### Internal` heading, since it's not something users or managers are likely to care much about.

app/page.tsx
```diff
@@ -162,7 +161,8 @@ export default async function Dashboard() {
 
       <div className="container mx-auto py-10">
         <MessageTable columns={experimentColumns} data={experimentAndBranchInfo} />
+        <a className="scroll-m-20 text-s" href="/api/auth/logout">Logout</a>
```
dmose, line 164: How about some whitespace above this to visually separate it from the table a bit?

dmose: I'm new to scroll margin and scroll snapping.  I'm curious what the intent here is -- I can't see any visual difference on my Mac if I disable it.

middleware.ts
```diff
@@ -0,0 +1,13 @@
+import { initAuth0, withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/edge';
+
+const isDevEnvironment = process.env.IS_DEV_ENV === 'true';
```
dmose, line 3: Please put IS_DEV_ENV into `env.sample` as well.

dmose:
In general, things are looking good.  I've just put a few nits here, and communicated some other documentation pieces in Slack.

middleware.ts
AllegroFox: sure!

CHANGELOG.md
AllegroFox: done!

app/page.tsx
AllegroFox: I think I added the scroll margin to match the table before I decided to put the link inside the <div>. It doesn't seem to have an effect now, so I've removed it. Adding padding to the <a> tag also doesn't seem to be having an effect however? I'm going to keep playing with this but I don't think we should block the PR on it :)

dmose:
Thank you for all your patience, persistence, and hard work -- we're almost there.

I think the main thing left to do here is for us to solidify/align around the documentation around how things are supposed to behave in different modes, so that: 

1. it's written down for people to check when they make changes
2. we can easily test that things are behaving correctly both now and later
3. I can think through the explicit requirements as I go through the code / configuration for the last time to convince myself that we haven't missed anything in our implementation

I also want to test the Set-Cookie stuff myself with the final code.

CONTRIBUTING.md
```diff
@@ -3,6 +3,64 @@
...
+
+Within the Auth0 dashboard, set the following properties:
+
+```
+Application Login URI: https://fxms-skylight.netlify.app/api/auth/login
+
+Allowed Callback URLs: http://localhost:3000/api/auth/callback
+
+Allowed Logout URLs: https://fxms-skylight.netlify.app/, http://localhost:3000
+```
```
dmose, line 60: It's not obvious to me why any of the local development callback URLs would points to Netlify. If this is truly necessary, can you clarify why in the text?

middleware.ts
```diff
@@ -0,0 +1,13 @@
+import { initAuth0, withMiddlewareAuthRequired } from '@auth0/nextjs-auth0/edge';
+
+const isDevEnvironment = process.env.NODE_ENV === 'development';
```
dmose, line 3: Based on https://nextjs.org/docs/messages/non-standard-node-env, we might want this for 'test' too.  Any idea why our tests are working without this?

CONTRIBUTING.md
```diff
@@ -3,6 +3,64 @@
...
+npm run start
+```
+to run the build, which simulates the prod environment.
+
+To test Auth0 locally, please create an account at https://auth0.com/, create a new app,
+and add the correct properties to `.env.local`:
+  
+Auth0 properties
+# A long, secret value used to encrypt the session cookie 
+# (use [openssl rand -hex 32] to generate a 32 bytes value)
```
dmose, line 37:
```suggestion
# (use `openssl rand -hex 32` to generate a 32 byte value)
```

CONTRIBUTING.md
```diff
@@ -3,6 +3,64 @@
...
+# The url of the Auth0 tenant domain
+AUTH0_ISSUER_BASE_URL='AUTH0_TENANT_URL'
+
+# The application's Client ID
+AUTH0_CLIENT_ID='YOUR_AUTH0_CLIENT_ID'
+
+# The application's Client Secret
+AUTH0_CLIENT_SECRET='YOUR_AUTH0_CLIENT_SECRET'
+
+Within the Auth0 dashboard, set the following properties:
```
dmose, line 52:
```suggestion
Within the Application URIs section of the Auth0 application Settings pane, set the following properties:
```

CONTRIBUTING.md
```diff
@@ -3,6 +3,64 @@
...
+Local development currently has Auth0 turned off.
+It’s possible to enable Auth0 locally (if desired) for testing, by setting the 
+correct Auth0 values in `.env.local`, and using
+
+```bash
+npm run start
+```
+to run the build, which simulates the prod environment.
+
+To test Auth0 locally, please create an account at https://auth0.com/, create a new app,
```
dmose, line 32:
```suggestion
To test Auth0 locally, start with these steps:

1. Create an account at https://auth0.com/ by clicking the "Login" button and following the prompts
2. Open the Applications section in the sidebar
3. create a new app
```

CONTRIBUTING.md
AllegroFox: So I just tested this && the only one that's actually required to point to `https://fxms-skylight.netlify.app/` is the login URI; Auth0 won't accept a non-https URL or a localhost for that. I can put that explanation in the doc :)

CONTRIBUTING.md
```diff
@@ -3,6 +3,68 @@
...
+npm run build && npm run start 
+```
+to run the build, which simulates the prod environment.
+
+To test Auth0 locally, start with these steps:
+
+1. Create an account at https://auth0.com/ by clicking the "Login" button and following the prompts
+2. Open the Applications section in the sidebar
+3. create a new app
+and add the correct properties to `.env.local`:
```
dmose, line 37: Please add something about creating a test user in the Auth0 UI.

.env.sample
```diff
@@ -2,3 +2,19 @@
 
 # Preview
 # EXPERIMENTER_API_CALL="experiments/?status=Preview&application=firefox-desktop"
+
+# Auth0 properties
+# A long, secret value used to encrypt the session cookie
```
dmose, line 7: Please add the thing from elsewhere in the PR about how to use `openssl` to generate this.

dmose:
Attempting to look at the deploy preview in this PR, I just get a page that says "secret is required".  Is that because we're in a weird state now, or should it just be showing the regular Skylight with Auth0?

app/page.tsx
```diff
@@ -160,9 +159,12 @@ export default async function Dashboard() {
         Total: {totalExperiments}
       </h5>
 
-      <div className="container mx-auto py-10">
+      <div className="space-y-5 container mx-auto py-10">
         <MessageTable columns={experimentColumns} data={experimentAndBranchInfo} />
+        <div>
+          <a className="text-s" href="/api/auth/logout">Logout</a>
+        </div>
```
dmose, line 166: This wants to be gated on NODE_ENV, I think

dmose:
This is looking good.  Once the comments here and in the deploy doc are addressed, I think we should be ready to go.  I want to give it one last look at that point.

dmose:
I looked at the Set-Cookie stuff, and I didn't see anything that appeared problematic either, so I think we're good on that front.

dmose (approved):
This looks good to me; thanks so much for chasing down the environment variable weirdness.

Ping me when you're ready to deploy...

r=dmose

PR #210, opened 2024-05-13
Bug 1896416 - Include Rollout messages in Skylight and add them to a separate table

sarahhjchung:
**Bug 1896416**

Changes made:
- Added a new table to show the current message rollout experiments
- Moved `isRolloutRecipe()`, `isExpRecipe()` and `isMsgRecipe()` to `NimbusRecipe` and added unit tests


dmose (changes requested):
This is looking quite nice; if I had realized how good this would look earlier, I would prioritized the bug more highly. Thanks for these quick feedback loops, Sarah!

Some thoughts:
* from a UX standpoint, this feels to me like a great step forward, because it makes the overview of messages-in-the-field more comprehensive, which is a key part of our initial value proposition.
* how about changing the rollout table header to "Current Message Rollouts" so that it matches the "Current Message Experiments" table?

I'll go ahead with some code suggestions as well:

Now that we're using `isRolloutRecipe()`, `isExpRecipe()` and `isMsgRecipe()` more, how about moving them onto NimbusRecipe and adding some unit tests?

After that, it should (fingers-crossed) be easy to fetch the complete message collection once, get a subset collection of messaging recipes once, and then filter that collection down to both an experiments-only- and a message-only-collection.

Thoughts?

sarahhjchung:
Yes, I can easily change the table name to match the other one! And I agree that moving the recipe functions to `NimbusRecipe` and adding unit tests would be good to do in this pr.

lib/nimbusRecipe.ts
```diff
@@ -250,4 +253,25 @@ null,
 
     return `https://experimenter.services.mozilla.com/nimbus/${encodeURIComponent(this._rawRecipe.slug)}/summary#${screenshotsAnchorId}`
   }
+
+  /**
+   * @returns true if this recipe is a messaging recipe.
+   */
+  isMsgRecipe() {
+    return this.usesMessagingFeatures()
+  }
```
sarahhjchung, line 262: I noticed that `isMsgRecipe()` is essentially just using `useMessagingFeatures()`. I didn't add unit tests because of this but wanted to ask if this function should then just be removed or if there might be future additions to this function which is why we're keeping it?

dmose: Good call, we can just remove `isMsgRecipe()`.

lib/nimbusRecipe.ts
```diff
@@ -250,4 +252,18 @@ null,
...
+  isExpRecipe() {
+    return !this._rawRecipe.isRollout
+  }
+
+  /**
+   * @returns true if this recipe is currently a message rollout experiment.
+   */
+  isMsgRolloutRecipe() {
+    return this.usesMessagingFeatures() && this._rawRecipe.isRollout
+  }
```
dmose, line 268: I'd prefer we keep the API simple and composable.  How about just having the callers use `usesMessagingFeatures()` and `!isExpRecipe()` and omitting this function?

components/ui/badge.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react"
```
dmose, line 1: We can remove this file from the PR now, correct?

app/page.tsx
```diff
@@ -53,43 +53,60 @@ async function getMsgExpRecipeCollection(): Promise<NimbusRecipeCollection> {
   await recipeCollection.fetchRecipes()
```
dmose, line 53: Instead of calling fetchRecipes twice, please hoist this call and the rollouts one to the Dashboard component, create a collection for everything and fetch once there, then make both getMsgExpRecipeCollection and getMessageRolloutRecipeCollection take a NimbusRecipeCollection as a parameter.

dmose (changes requested):
Looking good; the changes I'm proposing here should be pretty straightforward.

dmose:
Also, please add an entry to CHANGELOG.md describing this change.

__tests__/lib/nimbusRecipe.test.ts
```diff
@@ -250,4 +250,26 @@ describe('NimbusRecipe', () => {
...
+    it('returns true if the recipe is an experiment recipe not in rollout', () => {
+      const rawRecipe = ExperimentFakes.recipe("test-recipe");
+      const nimbusRecipe = new NimbusRecipe(rawRecipe)
+
+      const result = nimbusRecipe.isExpRecipe()
+
+      expect(result).toBe(true)
+    })
+
+    it('returns false if the recipe is a message rollout experiment', () => {
```
dmose, line 264:
```suggestion
    it('returns false if the recipe is a message rollout', () => {
```

app/page.tsx
```diff
@@ -47,49 +47,69 @@ async function getASRouterLocalMessageInfoFromFile(): Promise<FxMSMessageInfo[]>
...
-  return msgExpRecipeCollection
+  return msgRolloutRecipeCollection;
 }
 
 export default async function Dashboard() {
+  const recipeCollection = new NimbusRecipeCollection()
+  await recipeCollection.fetchRecipes()
+  console.log('recipeCollection.length = ', recipeCollection.recipes.length)
+
   // XXX await Promise.all for both loads concurrently
```
dmose, line 91: If you could change this comment to refer to Promise.allSettled, and file a bug to implement this speedup, that would be very helpful.

dmose (approved):
This looks great!  Just a couple of tiny tweaks here, and then it's ready to merge!  r=dmose

PR #213, opened 2024-05-15
Bug 1890870 - Make branches collapsible for experiment tables

sarahhjchung:
**Bug 1890870**

Changes made:
- Used the [Tanstack Table Expanding API](https://tanstack.com/table/v8/docs/guide/expanding) to make branch rows collapsible
- The chevron arrows by each experiment collapses/expands the branches to the specific experiment
- The chevron arrow at the header of the table collapses/expands all of the experiments all at once
- Currently, only toggling the chevron buttons (not the whole row) will collapse/expand

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Friday, May 17th, 2024
+
+### Added
+
+* Experiment rows now have toggle buttons that will expand/collapse all its branches
+* The experiments table header now has a toggle button that will expand/collapse all of the experiments at once
+
 ## Wednesday, April 24th, 2024
 
 ### Added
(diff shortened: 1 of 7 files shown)


dmose:
This looks slick!  I've added @AllegroFox to the list of possible reviewers, as I'm trying to unblock some other things right now, and I suspect she may be able to get to it sooner...

AllegroFox:
This looks great! One suggestion would be to give the hidden content some width, so that the table headers don't bounce around when we collapse/expand the rows. I'm looking at the Tanstack API now to see if there's a suggested way of doing that - ordinarily it would be something like using `hidden` instead of `display: none`, but since we're doing things with sub-rows instead there may be a better way.

dmose:
If you could add some MessageTable tests to verify the expand/collapse behaviors, I think that help avoid regressions as we evolve that code in the future.

AllegroFox (approved):
R+ with Dan's request for tests addressed; we can tweak the table/header layout in a followup!

dmose:
Also, please add a CHANGELOG entry.  Thanks!

sarahhjchung:
@AllegroFox I just re-requested for review because I added more changes than I expected to this PR. Thank you in advance!

AllegroFox (approved):
Looks great! Thank you for this 😃

PR #214, opened 2024-05-16
Display CTR data inline

dmose:
I've updated this to use environment variables for looker creds and added more documentation that I've run across...

The next step for this is to use it to break down into a user story, acceptance criteria, and likely a technical checklist.


app/page.tsx
```diff
@@ -87,36 +111,94 @@ async function getMsgRolloutCollection(
...
-  const localData = await getASRouterLocalMessageInfoFromFile()
-  const msgExpRecipeCollection = await getMsgExpRecipeCollection(recipeCollection)
-  const msgRolloutRecipeCollection = await getMsgRolloutCollection(recipeCollection)
+  const localData = await getASRouterLocalMessageInfoFromFile();
+  const msgExpRecipeCollection = await getMsgExpRecipeCollection(
+    recipeCollection
+  );
+  const msgRolloutRecipeCollection = await getMsgRolloutCollection(
+    recipeCollection
+  );
```
dmose, line 127: Looks like all the above changes to `Dashboard` are just formatting changes.  We should get Prettier landed before long so that this stops happening. We can probably leave these changes in the PR; this not is just here for PR readers...

app/page.tsx
```diff
@@ -11,21 +12,40 @@ import Link from "next/link";
...
     metrics: 'some metrics',
-    ctrPercent: .5, // getMeFromLooker
+    ctrPercent: 0, // getMeFromLooker
     ctrPercentChange: 2, // getMeFromLooker
     previewLink: getPreviewLink(messageDef),
   };
 
-  fxmsMsgInfo.ctrDashboardLink = getDashboard(messageDef.template, messageDef.id,
-    "release")
+  let dbElement = await getAWDashboardElement0();
```
dmose, line 28: We don't actually need this here at all.

app/page.tsx
```diff
@@ -11,21 +12,40 @@ import Link from "next/link";
...
+
+  // console.log("queryResult: ", queryResult)
+  if (queryResult.length > 0 && fxmsMsgInfo.template !== 'infobar') {
+    fxmsMsgInfo.ctrPercent = Number(Number(queryResult[0].primary_rate * 100).toFixed(1))
+  }
+  fxmsMsgInfo.ctrDashboardLink = getDashboard(messageDef.template, messageDef.id, "release")
+
+  // dashboard link -> dashboard id -> query id -> query -> ctr_percent_from_lastish_day
+
+  // console.log("fxmsMsgInfo: ", fxmsMsgInfo)
```
dmose, line 48: I think all the changed lines above should be hoisted into their own function, since we're effective copy-pasting this code three times inside this file.  Whether or not the function wants to be massaged so that it mostly lives in looker.ts, I'm not sure.  My guess is that there's not a lot of win to doing that right this moment, but maybe that depends on testability.

app/page.tsx
```diff
@@ -11,21 +12,40 @@ import Link from "next/link";
...
-  fxmsMsgInfo.ctrDashboardLink = getDashboard(messageDef.template, messageDef.id,
-    "release")
+  let dbElement = await getAWDashboardElement0();
+
+  // console.log("dbElement.query: ", dbElement.query)
+  // console.log("dbElement.query.filters: ", dbElement.query.filters)
+  // console.log("dbElement.query.model: ", dbElement.query.model)
+  // console.log("dbElement.query.filter_expression: ",
+  //   dbElement.query.filter_expression)
+
```
dmose, line 35: And since we don't have dbElement here, these can all go too.

dmose (commented):
It'll be interesting to look at this PR in light of the changes I'm suggesting here as well as the tests you've already written...

app/page.tsx
sarahhjchung: Yeah, that sounds good. I can file a bug for landing Prettier

app/page.tsx
sarahhjchung: For now I have the logic inside its own function and moved to `looker.ts` so that I can try testing it along with the other looker functions

app/page.tsx
```diff
@@ -87,36 +99,78 @@ async function getMsgRolloutCollection(
...
+        return updatedRecipe;
+      }
+    )
+  );
 
-  const totalExperiments = msgExpRecipeCollection.recipes.length
+  const totalExperiments = msgExpRecipeCollection.recipes.length;
 
-  const msgRolloutInfo: RecipeOrBranchInfo[] =
+  const msgRolloutInfo: RecipeOrBranchInfo[] = await Promise.all(
```
dmose, line 144: This looks identical(?) to the experiments one, so it should be hoistable into its own function too, I think.

dmose, line 144: This looks (almost) identical to the rollout one, so it can be hoisted into its own function too, further de-duping the code and making it more testable.

lib/looker.ts
```diff
@@ -0,0 +1,59 @@
...
+  // console.log(" newQuery result: ", result)
+  return result
+}
+
+export async function setCTRPercent(id: string, template?: string): Promise<number|undefined> {
+  const queryResult = await runEventCountQuery(
+    { 'event_counts.message_id':  '%' + id + '%' }
+  )
+  if (queryResult.length > 0 && template !== 'infobar') {
+    return Number(Number(queryResult[0].primary_rate * 100).toFixed(1))
```
dmose, line 57: I think the only reason it fails in infobar is because I didn't call it primary_rate there.  The API Explorer could help you figure this out. The best fix would be to fork the infobar dashboard and change the name in the new copy.

sarahhjchung: I also noticed that the data from today is included in the infobar dashboard, which shows 0 impressions which gives a 0% CTR. So currently investigating that 👀

dmose: If you haven't already done it, the InfoBar fix could be a separate bug/ticket so that it doesn't slow down this one.

dmose: I wonder if the infobar dashboard problem is another case-sensitivity thing

sarahhjchung: Yeah, I think I'll file a separate bug for this. I was able to make the dashboard show the last 30 complete days to remove today's incomplete data, but it was still giving a 0% CTR and I wasn't able to figure out why.

lib/looker.ts
```diff
@@ -0,0 +1,70 @@
...
+  newQueryBody.filters = Object.assign(
+    {
+      'event_counts.message_id': '%FAKESPOT_OPTIN_DEFAULT%',
+      'event_counts.submission_timestamp_date': '2 days'
+    },
+    filters
+  )
+
+  // console.log("filters: ", filters)
+  // console.log("newQueryBody.filters: ", newQueryBody.filters)
```
sarahhjchung, line 40: Should I leave these console.log statements here for debugging or can I remove these?

lib/nimbusRecipeCollection.ts
```diff
@@ -33,4 +36,38 @@ export class NimbusRecipeCollection implements NimbusRecipeCollectionType {
 
     return this.recipes
   }
+
+  /**
+   * @returns a list of RecipeInfo with updated ctrPercent properties for all
+   * the branches of every recipe in this collection
+   */
+  async getExperimentAndBranchInfos(): Promise<RecipeOrBranchInfo[]> {
```
sarahhjchung, line 44: I decided to move this function here because Next.js was complaining about multiple exports coming from `page.tsx` and I had the function exported for testing purposes.

sarahhjchung (commented):
I'll file follow up bugs for improving performance since we can keep Looker disabled until then. For now, I think this PR is ready for review and I just left some questions/comments here, thanks!

CONTRIBUTING.md
```diff
@@ -63,6 +63,13 @@ The Login URI must point at `https://fxms-skylight.netlify.app/api/auth/login`,
 
 Documentation for Auth0 quickstart can be found at https://auth0.com/docs/quickstart/webapp/nextjs
 
+## Enabling Looker Locally
+
+Local development by default has Looker disabled, via the `IS_LOOKER_ENABLED` env var. It’s possible to enable Looker locally (if desired) for testing, by setting `IS_LOOKER_ENABLED` to true && setting the correct Looker SDK credentials in `.env.local`.
+
+<!-- XXX link correct resource for devs to reach out to -->
+Please reach out to Data Engineering to retrieve Looker SDK credentials if needed.
```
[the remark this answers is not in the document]
dmose: I'd suggest replacing that with something like 'Ask in #data-help on Slack to get Looker API credentials for your own Looker account to use in development'

app/columns.tsx
```diff
@@ -35,8 +35,8 @@ export type FxMSMessageInfo = {
   template: string
   surface: string
   segment: string
-  ctrPercent: number
-  ctrPercentChange: number
+  ctrPercent?: number
+  ctrPercentChange?: number
```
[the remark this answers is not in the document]
dmose: That seems reasonable.  Please add a comment or two documenting why they are optional here.

lib/looker.ts
```diff
@@ -0,0 +1,70 @@
...
+  const origQuery = element0.query as IWriteQuery
+
+  // take the query from the original dashboard
+  const newQueryBody = structuredClone(origQuery)
+  delete newQueryBody.client_id // must be unique per-query
+
+  // override the filters
+  newQueryBody.filters = Object.assign(
+    {
+      'event_counts.message_id': '%FAKESPOT_OPTIN_DEFAULT%',
```
dmose, line 33: I did some testing a while back; it should be OK to remove this line.

lib/looker.ts
dmose: Feel free to remove them.  At some point we're going to want a real, more permanent logging strategy, but for now we might as well keep the code clean.

lib/looker.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @param template the message template
+ * @returns a CTR percent value for a message if the Looker query results are
+ * defined
+ */
+export async function getCTRPercent(id: string, template?: string): Promise<number|undefined> {
+  const queryResult = await runEventCountQuery(
+    { 'event_counts.message_id':  '%' + id + '%' }
+  )
+
+  // XXX fix issues with the infobar Looker dashboards to remove the
```
dmose, line 65: Link to the bug for this in the comment, please.

lib/nimbusRecipeCollection.ts
dmose: Very reasonable.

lib/looker.ts
```diff
@@ -0,0 +1,70 @@
...
+  const dashboardId = "1471";
+//  await(getDBFilters())
+
+  const elements: IDashboardElement[] = await SDK.ok(SDK.dashboard_dashboard_elements(dashboardId));
+  return elements[0];
+}
+
+export async function runEventCountQuery(filters: any): Promise<any>{
+  const SDK = LookerNodeSDK.init40()
+  const me = await SDK.ok(SDK.me());
```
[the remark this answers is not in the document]
dmose: Your concern is valid. How about moving the `LookerNodeSDK.init40()` line and assignment to the top-level, where it will be executed and module load time, and getting rid of the `SDK.me` line entirely, since we don't use the result...

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Friday, June 7th, 2024
+
+### Added
+
+* When enabling Looker, CTR percents will be displayed in the Metrics column inside tables
```
dmose, line 5:
```suggestion
* Display CTR percentages in the table when Looker is configured
```

app/columns.tsx
```diff
@@ -139,8 +141,10 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
         return ( <></> );
       }
 
-      if (props.row.original.ctrDashboardLink) {
-        return OffsiteLink(props.row.original.ctrDashboardLink, "Dashboard");
+      if (props.row.original.ctrDashboardLink && props.row.original.ctrPercent !== undefined && props.row.original.template !== 'infobar') {
+        return OffsiteLink(props.row.original.ctrDashboardLink, props.row.original.ctrPercent + "% CTR");
+      } else if (props.row.original.ctrDashboardLink) {
+        return OffsiteLink(props.row.original.ctrDashboardLink, "Dashboard")
```
dmose, line 147: The above clause looks identical to the next one except for the exclusion of infobar CTRs.  If they're broken at the moment, don't we want to exclude them both places?

If so, I'd suggest hoisting this code into its own function. Please also include a link to the bug about enabling inline CTR percentages for infobar.

app/columns.tsx
```diff
@@ -302,8 +306,10 @@ export const experimentColumns: ColumnDef<RecipeOrBranchInfo>[] = [
 
       // XXX see https://bugzilla.mozilla.org/show_bug.cgi?id=1890055 for
       // re-enabling infobar code.
```
dmose, line 308: Ironically, this comment is obsolete because it refers to a different thing where infobars were entirely disabled.  Let's remove it now.

app/page.tsx
```diff
@@ -1,31 +1,43 @@
...
     product: 'Desktop',
     id: messageDef.id,
     template: messageDef.template,
     surface: getDisplayNameForTemplate(getTemplateFromMessage(messageDef)),
     segment: 'some segment',
     metrics: 'some metrics',
-    ctrPercent: .5, // getMeFromLooker
-    ctrPercentChange: 2, // getMeFromLooker
+    ctrPercent: undefined, // getMeFromLooker
+    ctrPercentChange: undefined, // getMeFromLooker
```
dmose, line 25: Please change the `// getMeFromLooker` in these lines to `// may be populated from Looker data`.

lib/looker.ts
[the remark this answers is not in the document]
sarahhjchung: When removing `SDK.me` and moving the `LookerNodeSDK.init40()` line to the top-level, there are still build failures that show this error:
```
Error: TS-SDK configuration error:
Missing required configuration values like base_url
```
If we are trying to use the service account in prod, does it make sense to add the `IS_LOOKER_ENABLED` and `LOOKERSDK_BASE_URL` env variables to the `.env` file in order to avoid these build failures?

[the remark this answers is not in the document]
dmose: This is definitely the right question.  Generally, this doesn't seem like a great idea, but I'm wondering if the is going to end up being yes in this case, since we want to pre-render the web pages.  I'm reading through some server components and poking around a bit as I think about this.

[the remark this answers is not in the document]
dmose: So I spent a bunch of time this afternoon wrapping my head around the many possible ways to structure NextJS 14 static vs dynamic pages and how that interleaves with caching, and I will start by saying that I'm very glad the NextJS 15 docs talk a bunch about how they're going to keep working to make this better.  The complexity of the developer experience is currently.... painful.

In any case, my biggest conclusion is that we don't want to block landing this PR on getting all the caching stuff sorted out. 

This makes me think that my suggestion of hoisting the SDK initialization isn't the right way to go.  Since the init40 method isn't obviously idempotent, I'd suggest we move the SDK initialization stuff to a method in looker.ts that looks something like this:

```
let SDK;

function getLookerSDK() {
	if (!SDK) {
		SDK = LookerNodeSDK.init40()
	}
	return SDK
}
```

More Tuesday, though I'm not quite sure when, I have some meetings and need to prioritize Road Show work over the CTR stuff, unforch.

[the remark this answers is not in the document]
sarahhjchung: Ahh I see. I'll move the SDK initialization as per your suggestion for now and hopefully we can find some time to pair soon!

__tests__/app/message-table.test.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { MessageTable } from "@/app/message-table";
-import { experimentColumns, RecipeOrBranchInfo } from "@/app/columns";
+import { experimentColumns, fxmsMessageColumns, FxMSMessageInfo, RecipeInfo, RecipeOrBranchInfo } from "@/app/columns";
 import { ExperimentFakes } from "@/__tests__/ExperimentFakes.mjs";
+import { NimbusRecipeCollection } from "@/lib/nimbusRecipeCollection";
 import { NimbusRecipe } from "@/lib/nimbusRecipe";
 
+jest.mock("../../lib/sdk")
+
+global.structuredClone = jest.fn((val) => {
```
dmose, line 10: This seems to be an issues with jsdom.  To avoid having to define this all over the place, try using something like the solution shown in https://github.com/jsdom/jsdom/issues/3363#issuecomment-1575996129.

dmose (commented):
More review comments to come later today or tomorrow.

.env
```diff
@@ -8,3 +8,7 @@ EXPERIMENTER_API_PREFIX="https://experimenter.services.mozilla.com/api/v7/"
 # https://htmlpreview.github.io/?https://github.com/mozilla/experimenter/blob/main/docs/experimenter/swagger-ui.html has more info.
 #
 EXPERIMENTER_API_CALL="experiments/?status=Live&application=firefox-desktop"
+
+# Looker configurables
+IS_LOOKER_ENABLED=false
+LOOKERSDK_BASE_URL=null
```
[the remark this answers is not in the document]
sarahhjchung: Yeah, because I call `getLookerSDK()` in the `sdk.ts` file, the build complains about having no base URL defined even when `IS_LOOKER_ENABLED=false`.

__tests__/app/message-table.test.tsx
dmose: Thanks

lib/looker.ts
```diff
@@ -0,0 +1,53 @@
+import { IDashboardElement, IWriteQuery } from "@looker/sdk"
+import { SDK } from "./sdk";
+
+export async function getAWDashboardElement0(): Promise<IDashboardElement> {
+  const dashboardId = "1471";
```
dmose, line 5: Don't necessarily need to fix this now, since we've spun off a bug, but this is why the Infobar CTR stuff isn't working, in part.  We have a different dashboard ID for infobars (in addition to have a different JSON property name or two in the results).

lib/looker.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const elements: IDashboardElement[] = await SDK.ok(SDK.dashboard_dashboard_elements(dashboardId));
```
dmose, line 7: Add an XXX comment here for the performance work: since we're only getting one element, we should be able to switch this out for the more performant `dashboard_element`.  https://mozilla.cloud.looker.com/extensions/marketplace_extension_api_explorer::api-explorer/4.0/methods/Dashboard/dashboard_element has more info.

lib/nimbusRecipeCollection.ts
```diff
@@ -33,4 +36,38 @@ export class NimbusRecipeCollection implements NimbusRecipeCollectionType {
...
+   * ctrPercent properties
+   */
+  async getExperimentAndBranchInfos(): Promise<RecipeOrBranchInfo[]> {
+    return await Promise.all(
+      this.recipes.map(
+        async (recipe: NimbusRecipe): Promise<RecipeInfo> => {
+          let updatedRecipe = recipe.getRecipeInfo();
+          
+          // Update all branches with CTR data for the recipe
+          updatedRecipe.branches = await Promise.all(
```
dmose, line 51: For even better readability, please pull this out into its own method also.

lib/nimbusRecipeCollection.ts
```diff
@@ -1,12 +1,37 @@
...
   fetchRecipes: () => Promise<Array<NimbusRecipe>>
 }
 
+/**
+ * @returns an array of BranchInfo with updated CTR percents for the recipe
+ */
+async function updateBranchesCTR(recipe: NimbusRecipe): Promise<BranchInfo[]> {
+  return await Promise.all(
+    recipe.getBranchInfos().map(
+      async (branch: BranchInfo): Promise<BranchInfo> => {
```
dmose, line 19: Please rename this to branchInfo, since there's already a thing called branches inside nimbus recipes that is different.

app/columns.tsx
```diff
@@ -62,7 +64,7 @@ export type RecipeInfo = {
   nimbusExperiment: NimbusExperiment
   isBranch?: boolean
   branches: BranchInfo[]
```
dmose, line 66: Please add something like 
```
// XXX rename this to branchInfos to avoid confusion with the thing called branches inside nimbus recipes themselves
```

dmose (commented):
This is looking great; we're almost there!  If you could address the minor stuff I've called out here and then merge in the latest stuff from `main`, I'd like to have a look at that, and then I'm hoping to stamp it with r+.

Thanks for you patience!

lib/__mocks__/sdk.ts
```diff
@@ -0,0 +1,47 @@
...
+}];
+
+export function getLookerSDK(): any {
+    return "mocked SDK"
+}
+
+export const SDK = {
+    dashboard_dashboard_elements: () => "dashboard_dashboard_elements",
+    create_query: () => "create_query",
+    run_query: () => "run_query",
```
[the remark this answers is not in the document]
sarahhjchung: Yes you're right, that makes it a lot clearer

ctr-spike.md
```diff
@@ -0,0 +1,63 @@
...
+
+  * running the queries in parallel might have complexity issues (doesn't seem super high risk).
+  * Dependency: Finish sorting out how to [get a limit privs service account](https://cloud.google.com/looker/docs/api-auth#authentication_with_an_sdk) for the web app to use the Looker API on our instance. I don't yet understand the model for constraining the privs for that account.
+    * [DSRE-1583](https://mozilla-hub.atlassian.net/browse/DSRE-1583) (**DONE**)
+  * Dependency: add getting secrets from environment vars
+    * [Bug 1888930](https://bugzilla.mozilla.org/show_bug.cgi?id=1888930) (**DONE**)
+  * Performance issues
+    * do stuff in parallel as described above
+
+### BigQuery
```
dmose, line 55: I'd suggest deleting this whole section, as it's no longer relevant.

dmose (commented):
This looks great; thanks for all your hard work.  r=dmose with that one nit fixed.  Github won't let me approve this since I was the one who originally started the pull request, so we'll need @AllegroFox to do that...

sarahhjchung (approved):

PR #220, opened 2024-05-22
Bug 1879140 - Fix observatory.mozilla.org issues

sarahhjchung:
[**Bug 1879140**](https://bugzilla.mozilla.org/show_bug.cgi?id=1879140)

Changes made:
- Added a X-Frame-Options header
- Added a Content-Security-Policy header for both dev and prod
- Added a [Content-Security-Policy-Report-Only](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy-Report-Only) header to test the prod csp header before applying the header 
- [Observatory test results for this PR deploy preview](https://observatory.mozilla.org/analyze/deploy-preview-220--fxms-skylight.netlify.app) (NOTE: currently the CSP test is failing because I have the header for it commented out and the report-only header replacing it due to reasons explained in my comment below)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,3 +3,9 @@
 
   [plugins.inputs.settings]
     preset = "desktop" # Optionally run Lighthouse using a desktop configuration
+
+[[headers]]
+  for = "/*"
+  [headers.values]
+    Content-Security-Policy: "frame-ancestors 'none'"
+    X-Frame-Options = "DENY"
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,6 +7,5 @@
 [[headers]]
   for = "/*"
   [headers.values]
-    // Content-Security-Policy-Report-Only = "default-src https:; report-to /csp-violation-report-endpoint/"
     Content-Security-Policy = "frame-ancestors 'none'"
     X-Frame-Options = "DENY"
--- a/netlify.toml
+++ b/netlify.toml
@@ -7,5 +7,6 @@
 [[headers]]
   for = "/*"
   [headers.values]
-    Content-Security-Policy: "frame-ancestors 'none'"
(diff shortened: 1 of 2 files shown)


sarahhjchung:
When testing the CSP headers locally (`cspHeaderDev`) and in the deploy preview (`cspHeaderProd`), I was able to pass all the observatory tests and see no console errors. Currently, I have commented out the CSP header and added a `Content-Security-Policy-Report-Only` header so I can test the `cspHeaderProd` in prod without breaking it. If this is okay, I can create a new PR to remove the `Content-Security-Policy-Report-Only` header after ensuring everything looks good in prod.

AllegroFox:
Hi Sara! I've tested this locally and I think we're safe to land it. I also rebased it on main so I could try it out with the Auth0 stuff and I don't see any issues, though we'll probably want to run everything through Observatory again anyway since the Auth stuff will introduce cookies as well. 

I think your strategy of merging without the CSP to check for issues is a good one, let's go ahead and do that and then file any followups we need!

If you request review from me I can give R+ :)

sarahhjchung:
> Hi Sara! I've tested this locally and I think we're safe to land it. I also rebased it on main so I could try it out with the Auth0 stuff and I don't see any issues, though we'll probably want to run everything through Observatory again anyway since the Auth stuff will introduce cookies as well.
> 
> I think your strategy of merging without the CSP to check for issues is a good one, let's go ahead and do that and then file any followups we need!
> 
> If you request review from me I can give R+ :)

Thanks Emily! I'll run the Observatory tests again after rebasing on my end, then I'll merge it with the CSP report only!

AllegroFox (approved):
Thanks! R+

PR #223, opened 2024-05-24
Enable preview links for Feature Callout

AllegroFox:
This enables preview links for Feature Callout messages both in the prod table && the experiments table. It also fixes a bug where about:welcome messages that had empty value objects were showing broken previews.

https://bugzilla.mozilla.org/show_bug.cgi?id=1898789 Is doing some work on the Firefox side of things to ensure the feature callouts display correctly; by manipulating the targeting, trigger && anchor point to be values that are present on about:messagepreview.

We could wait to land this until that bug makes it to release, or we could update the caveat on the Skylight page to be `a Firefox 128 build from May...` 

~~[WIP, just wanted it out of my brain for Friday]~~


dmose:
This is exciting -- thanks for putting it together!

dmose:
@sarahhjchung's going to have a look at this first, and then I'll have a look on Friday (I'm mostly on PTO Thursday)

lib/nimbusRecipe.ts
```diff
@@ -79,33 +79,45 @@ export class NimbusRecipe implements NimbusRecipeType {
...
         // XXX should iterate over all screens
         branchInfo.id = feature.value.content.screens[0].id
+        // Localize the feature callout if necessary
+        console.log(this._rawRecipe);
+        if(this._rawRecipe.localizations) {
+          console.log(Object.keys(this._rawRecipe.localizations)[0]);
+        }
+        let localizedFeatureCallout = _substituteLocalizations(feature.value,
+this._rawRecipe.localizations?.[Object.keys(this._rawRecipe.localizations)[0]])
+        //console.log(JSON.stringify(localizedFeatureCallout, null, 2));
```
sarahhjchung, line 118: Just to clean up the code, perhaps this comment can be removed:
```suggestion
```

app/columns.tsx
```diff
@@ -153,9 +153,9 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
       </div>
     ),
     cell: (props: any) => {
```
sarahhjchung, line 155: I think it might be useful to add some tests in `message-table.test.tsx` to test the correct rendering of the preview buttons for these message surfaces

sarahhjchung (commented):
I don't have many suggestions but this was a good PR to help introduce me to how message previews work in Skylight, so thanks!

lib/nimbusRecipe.ts
AllegroFox: Oh yep, good catch!

app/columns.tsx
AllegroFox: that's a good idea, I'll add some cases.

sarahhjchung (approved):
Thanks, looks good to me! But I think it'll be good to wait for Dan to have a final pass at this.

__tests__/app/message-table.test.tsx
```diff
@@ -85,5 +85,41 @@ describe("MessageTable", () => {
...
+      ];
+      render(
+        <MessageTable columns={experimentColumns} data={messageTableData} />
+      );
+
+      const toggleButton = screen.getByTestId("toggleAllRowsButton");
+      fireEvent.click(toggleButton);
+
+      expect(screen.getAllByText("Copy Preview URL").length).toBe(3);
+    });
```
dmose, line 123: This is complicated to read, and has the disadvantage that having the wrong two things broken at the same time could cancel each other out and show a false positive.

How about splitting this up?  I suspect that just doing it in four tests by hand would be the fastest...

dmose:
Some stuff I noticed while testing:

a. Often, the callout points to something that isn't actually there (e.g., PDF.js affordance, an icon in the address bar, etc.). I'm guessing this isn't trivial to fix. If that's true, please file a spin-off bug and link to it here.
b. the close button doesn't seem to work on anything except the last screen of a multi-screen callout. Please either fix this or if it's non-trivial, file a spin-off bug and link to it here.

AllegroFox:
Just making note of what we discussed in standup -
For the two cases you mentioned, Dan, both of these things are issues that come from Feature Callout itself. For a, it's because the message uses an anchor like `hbox#browser` which will always exist, but the message "in the wild" used other context to ensure the "real" anchor element was present. This should go away as we start to use in-content callouts.
For b, turns out it's because these callouts (mostly feature tour) are using a pref to determine whether to show or hide, and the dismiss button modifies the pref. We should be able to fix this by clearing the pref before rendering the message, but that work will have to be done on the Firefox side, coming from about:messagepreview - I can file a bug for it :)

dmose:
> We should be able to fix this by clearing the pref before rendering the message, but that work will have to be done on the Firefox side, coming from about:messagepreview - I can file a bug for it :)

That would be great; and please put a link to the bug here.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+
+## Wednesday, June 12, 2024
+
+### Added
+
+* Enabled the preview button for Feature Callouts
+
+### Fixed
+
+* Fixed a bug where about:welcome previews were being generated for empty messages
```
dmose, line 10:
```suggestion
* Fixed a bug where about:welcome previews were being generated for empty messages (those without a `screens` property)
```

__tests__/app/message-table.test.tsx
```diff
@@ -85,5 +85,90 @@ describe("MessageTable", () => {
...
+        nimbusRecipe.getRecipeInfo(),
+      ];
+      render(
+        <MessageTable columns={experimentColumns} data={messageTableData} />
+      );
+
+      const toggleButton = screen.getByTestId("toggleAllRowsButton");
+      fireEvent.click(toggleButton);
+
+      expect(screen.getAllByText("Copy Preview URL").length).toBe(1);
```
dmose, line 107: I like the style of having a one line table and counting the buttons!

__tests__/app/message-table.test.tsx
dmose: This looks good.  One more easy thing that will make them even easier to read/maintain: split each test into three chunks (setup, execute, verify) separated by whitespace.  For example:

```typescript
    it("does not render a preview button for toasts", () => {
      const toastRecipe = ExperimentFakes.recipe("test-recipe");
      // we don't generate a previewLink for toasts
      toastRecipe.branches[1].features[0].value.template = "toast_notification"
      toastRecipe.branches[1].features[0].value.content = { tag: "test tag" }
      const nimbusRecipe = new NimbusRecipe(toastRecipe);
      const messageTableData: RecipeOrBranchInfo[] = [
        nimbusRecipe.getRecipeInfo(),
      ];
      render(
        <MessageTable columns={experimentColumns} data={messageTableData} />
      );
      const toggleButton = screen.getByTestId("toggleAllRowsButton");

      fireEvent.click(toggleButton);

      const previewButton = screen.queryByText("Copy Preview URL");
      expect(previewButton).not.toBeInTheDocument();
    });
```

This is a lighter-weight version of what https://github.com/bencompton/jest-cucumber does, which would probably be even more readable.

dmose (approved):
This looks good; just a couple of nits and then it'll be ready to land.  r=dmose with nits addressed.

AllegroFox:
Feature tour pref will be addressed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1903643

PR #227, opened 2024-05-31
Remove the logout button since logout is not implemented

AllegroFox:
We never fully implemented the logout route in out Auth0, so the logout button doesn't do anything. This just removes it.


sarahhjchung (approved):
Thanks!

PR #228, opened 2024-05-31
Bug 1900121 - Add about:welcome defaults to message array

AllegroFox:
Adds the AboutWelcomeDefaults message to the release message array. This can land before the Feature Callout stuff since it's simpler, and doesn't depend on a Firefox patch.


lib/messageUtils.ts
```diff
@@ -76,7 +78,29 @@ export function toBinary(string: string): string {
...
+    defaultWelcomeFake.content.modal = "tab";
+    // The recipe might have a backdrop, but if not, fall back to the default
+    defaultWelcomeFake.content.backdrop = message.backdrop ||
+    "var(--mr-welcome-background-color) var(--mr-welcome-background-gradient)";
+
+    return defaultWelcomeFake;
+  }
+  // if the message isn't about:welcome, just send it back
+  return message;
+}
```
dmose, line 100: Ideally, we'd have a worthwhile test for this, but I think that to do that, we'd need to set up integration testing with a real Firefox build, and that's kinda out of scope here.

app/columns.tsx
```diff
@@ -153,8 +153,8 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
       </div>
     ),
     cell: (props: any) => {
-      if (props.row.original.template !== 'infobar'
-          && props.row.original.template !== 'spotlight') {
+      const supportedTypes = ['infobar', 'spotlight', 'defaultaboutwelcome'];
```
dmose, line 156: So it strikes me that (likely outside of this PR) we'll want to reuse this same dashboard/preview for control branches of about:welcome experiments. Do you agree? If so, is this `defaultaboutwelcome` feature the approach we want to use?

Whatever we do there, I don't think we should let that work hold up this PR landing unless it's super easy, so I suspect the thing to do is spin off a new bug for that and link to it here.

But I'm very interested in your thoughts here...

dmose (approved):
This looks great; thanks for making it happen!  I do have a question and comment, but r=dmose unless you make substantial changes and want me to look at it again. That said, you could also ask @sarahhjchung if you end up wanting another look before I have a chance to do so.

app/columns.tsx
AllegroFox: I think that could make sense! I would be inclined to spin it off into a spike because there's a bunch of neat stuff we could do - I was going to file a followup anyway to decide what to do about the screen targeting of the default message, and this could have an effect here depending on what we decide.

- we could pull in the defaultaboutwelcome object for control branches and leave it untouched, so it would display itself according to the state of the previewer's Firefox (that's what it's doing in this patch)
- We could strip screen targeting from the object so that *all* the possible screens show
- we could try and glean some kind of likely state from the relevant experiment targeting (i.e. locale, isDefault, isPinned) and then tweak the aboutwelcome object to show what a *user* would (likely) see as control
- we could go to 11 and build some kind of UX interface that would allow targeting to be tweaked/imitated for the preview (i.e. allow the previewer to specify what they want to see)

I think we should do that last one eventually anyway, but it will require some planning. As far as picking between the first three options in the interim, I'd like to maybe do a quick informal poll of some of the folks who are our prospective audience and see what they would find most useful, and then just do that.

PR #235, opened 2024-06-05
Quick hack to show completed experiments ordered by date

dmose:
This turned out to be useful in understanding what FxMS experiments were live at the time, so leaving it around for now.  Before too long, we'll want to build sorting and status filters into the UI so we don't need to hack up stuff like this.


dmose:
This has now been implemented in production without needing to do a special build; thanks @sarahhjchung!

PR #248, opened 2024-06-14
Bug 1902424 - Fix weird feature callout dashboard lossage

sarahhjchung:
[**Bug 1902424**](https://bugzilla.mozilla.org/show_bug.cgi?id=1902424)

Changes made:
- I added a fix for the weird branch ids for the following experiments: [Discovery of Search SAP Early Day EN](https://experimenter.services.mozilla.com/nimbus/discovery-of-search-sap-early-day-en/summary#treatment-b) and [Discovery of Search SAP Existing EN](https://experimenter.services.mozilla.com/nimbus/discovery-of-search-sap-existing-en/summary#treatment-b)
    - Those branch ids in the jsons had `:treatment-a` at the end despite it being treatment-b branches

### DIFF
--- a/lib/nimbusRecipe.ts
+++ b/lib/nimbusRecipe.ts
@@ -105,7 +105,9 @@ export class NimbusRecipe implements NimbusRecipeType {
 
       case 'feature_callout':
         // XXX should iterate over all screens
-        branchInfo.id = feature.value.content.screens[0].id
+        // XXX some branches have incorrect ":treatment-a" attached to the end
+        // of the id that needs to be removed (see https://bugzilla.mozilla.org/show_bug.cgi?id=1902424)
+        branchInfo.id = feature.value.content.screens[0].id.split(":")[0]
         break
 
       case 'infobar':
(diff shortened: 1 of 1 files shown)


lib/nimbusRecipe.ts
```diff
@@ -105,7 +105,9 @@ export class NimbusRecipe implements NimbusRecipeType {
 
       case 'feature_callout':
         // XXX should iterate over all screens
-        branchInfo.id = feature.value.content.screens[0].id
+        // XXX some branches have incorrect ":treatment-a" attached to the end
+        // of the id that needs to be removed (see https://bugzilla.mozilla.org/show_bug.cgi?id=1902424)
+        branchInfo.id = feature.value.content.screens[0].id.split(":")[0]
```
dmose, line 110: It would be good to expand the comment here on why this is an appropriate fix, especially and if there's something else that we'll want to consider for the longer term to prevent this from happening at all. As an example, are we depend on multiple JSON properties to be kept in sync by hand?  If so, this could be a candidate for linting (or maybe even JSON schema), or, someday, elimination of the error via GUI message creation.

dmose (approved):

lib/nimbusRecipe.ts
sarahhjchung: Ahh, I see. I think due to other experiments having lower-case message ids which Looker won't accept, there's already logic implemented to uppercase all message ids in the dashboard links. So for now, I may keep the trimming solution to avoid having to change that.

PR #256, opened 2024-06-20
Bug 1901036 - Display more accurate CTR percents including infobar messages

sarahhjchung:
[**Bug 1901036**](https://bugzilla.mozilla.org/show_bug.cgi?id=1901036)

Changes made:
- Duplicated the [infobar Looker dashboard](https://mozilla.cloud.looker.com/dashboards/1775?Messaging+System+Ping+Type=infobar&Submission+Date=30+day+ago+for+30+day&Messaging+System+Message+Id=INFOBAR_LAUNCH_ON_LOGIN_FINAL&Normalized+Channel=release&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=&Experiment=&Experiment+Branch=) and enabled infobar CTRs in Skylight
- I noticed the CTR percents displayed in Skylight didn't accurately reflect the CTR percents in the dashboards. So the looker API query filters have been updated to match the same filters in the dashboard links, which now makes the displayed CTR percents match the ones in the dashboards. 

Note: Looker needs to be enabled locally to test this patch

### DIFF
--- a/__tests__/lib/messageUtils.test.ts
+++ b/__tests__/lib/messageUtils.test.ts
@@ -1,4 +1,4 @@
-import { getDashboard, _isAboutWelcomeTemplate, toBinary } from '@/lib/messageUtils'
+import { getDashboard, _isAboutWelcomeTemplate, toBinary, getDashboardIdForTemplate } from '@/lib/messageUtils'
 
 describe('isAboutWelcomeTemplate', () => {
   it('returns true if a feature_callout', () => {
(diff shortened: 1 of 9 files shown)


__tests__/lib/messageUtils.test.ts
```diff
@@ -67,7 +67,7 @@ describe('getDashboard', () => {
 
     const result = getDashboard(template, msgId, channel, experiment, branchSlug)
 
-    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/1682?Messaging+System+Ping+Type=${encodeURIComponent(template)}&Submission+Date=30+days&Messaging+System+Message+Id=${encodeURIComponent(msgId)}&Normalized+Channel=${encodeURIComponent(channel)}&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=&Experiment=${encodeURIComponent(experiment)}&Experiment+Branch=${encodeURIComponent(branchSlug)}`
+    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/1775?Messaging+System+Ping+Type=${encodeURIComponent(template)}&Submission+Date=30+day+ago+for+30+day&Messaging+System+Message+Id=${encodeURIComponent(msgId)}&Normalized+Channel=${encodeURIComponent(channel)}&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=&Experiment=${encodeURIComponent(experiment)}&Experiment+Branch=${encodeURIComponent(branchSlug)}`
```
dmose, line 70: Decide what, if anything, needs to be dried up here.

lib/looker.ts
```diff
@@ -11,21 +14,30 @@ export async function getAWDashboardElement0(): Promise<IDashboardElement> {
   return elements[0];
 }
 
-export async function runEventCountQuery(filters: any): Promise<any>{
-  const element0 = await getAWDashboardElement0()
+export async function runEventCountQuery(filters: any, template: string): Promise<any>{
```
dmose, line 17: Since the infobar query is not run against the event count look, let's change this function's name to `runQueryForTemplate`. I would also make the template the first parameter, as it's the most important thing.

lib/looker.ts
```diff
@@ -40,17 +52,43 @@ export async function runEventCountQuery(filters: any): Promise<any>{
...
+): Promise<number | undefined> {
+  let queryResult;
+  if (template === "infobar") {
+    queryResult = await runEventCountQuery(
+      {
+        "messaging_system.metrics__text2__messaging_system_message_id": id,
+        "messaging_system.normalized_channel": channel,
+        "messaging_system.metrics__string__messaging_system_ping_type":
+          template,
+      },
```
dmose, line 76: I would expect the above list of filters to also include an experiment and a branch. Is there some reason there shouldn't be?

app/page.tsx
```diff
@@ -27,14 +27,16 @@ async function getASRouterLocalColumnFromJSON(messageDef: any) : Promise<FxMSMes
     previewLink: getPreviewLink(maybeCreateWelcomePreview(messageDef)),
   };
 
+  const channel = "release";
```
dmose, line 30: Nice cleanup here.

dmose (commented):
I still need to play with this some more, but here are some thoughts...

lib/looker.ts
sarahhjchung: Yes, this should include those too I missed them here!

dmose (approved):
This looks great. All of my comments are minor, so r=dmose with the comments addressed.  In addition to the changes inline in this PR, here are some comments on the updated dashboard:

* I like the fix to not show the last day without data.
* Please bump the version in the top-level title from v2 to v3
* Please change visual label from Primary Rate to Primary Button Clickthrough Rate.  If you do this via Visulization > Edit > Series > Customizations, it _should_ leave the lower level stuff still named primary_rate so that it's the same across both dashboards.
*

lib/nimbusRecipeCollection.ts
```diff
@@ -21,7 +21,10 @@ async function updateBranchesCTR(recipe: NimbusRecipe): Promise<BranchInfo[]> {
         // Looker being case sensitive
         const ctrPercent = await getCTRPercent(
           branchInfo.id.toUpperCase(),
-          branchInfo.template
+          branchInfo.template!,
+          undefined,
```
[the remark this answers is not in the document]
sarahhjchung: Agreed, I'll leave a comment. Also, I left the channel undefined because that's how it is when the dashboard link is being defined for branches (see [here](https://github.com/mozilla/skylight/blob/main/lib/nimbusRecipe.ts#L185-L186)).

PR #259, opened 2024-06-21
Bug 1903843 - Revise chevron icons

sarahhjchung:
[**Bug 1903843**](https://bugzilla.mozilla.org/show_bug.cgi?id=1903843)

Changes made:
- Revised the chevron icons for expanding/collapsing branches as per the UX feedback. 
- The icons are now:
    - Collapsed state = chevron right
    - Expanded state = chevron down
    - Expand all rows button = chevron up down (I am open to changing this to be just chevron right to keep it consistent with the other rows)


AllegroFox (approved):
Thanks for this Sarah! I think it's definitely clearer. I'm also a fan of the double arrow for the expand button. R+!

PR #266, opened 2024-06-26
Bug 1904388 - Constrain dashboard to the experiment length

sarahhjchung:
[**Bug 1904388**](https://bugzilla.mozilla.org/show_bug.cgi?id=1904388)

Changes made:
- Constrained dashboard link to filter dates based on experiment `startDate` and `endDate`
- The dashboard are restricted for the following cases:
    - Live message dashboards show data from the last 30 days (as it already was)
    - Experiment dashboards show data from `startDate` to today (since the `endDate` properties are always null)

### DIFF
--- a/lib/messageUtils.ts
+++ b/lib/messageUtils.ts
@@ -58,14 +58,16 @@ export function getDashboard(
   const encodedChannel = channel ? (encodeURIComponent(channel)) : "";
   const encodedExperiment = experiment ? (encodeURIComponent(experiment)) : "";
   const encodedBranchSlug = branchSlug ? (encodeURIComponent(branchSlug)) : "";
+  const encodedStartDate = startDate ? (encodeURIComponent(startDate)) : "";
+  const encodedEndDate = endDate ? (encodeURIComponent(endDate)) : "";
   const dashboardId = getDashboardIdForTemplate(template);
 
   // Showing the last 30 complete days to ensure the dashboard isn't including today which has no data yet
   let encodedSubmissionDate = "30+day+ago+for+30+day";
   if (startDate && endDate && (new Date() < new Date(endDate))) {
-    encodedSubmissionDate = `${startDate}+to+${endDate}`;
+    encodedSubmissionDate = `${encodedStartDate}+to+${encodedEndDate}`;
   } else if (startDate) {
-    encodedSubmissionDate = `${startDate}+to+today`;
+    encodedSubmissionDate = `${encodedStartDate}+to+today`;
   }
 
   if (_isAboutWelcomeTemplate(template)) {
--- a/lib/messageUtils.ts
+++ b/lib/messageUtils.ts
@@ -60,13 +60,10 @@ export function getDashboard(
   const encodedBranchSlug = branchSlug ? (encodeURIComponent(branchSlug)) : "";
   const dashboardId = getDashboardIdForTemplate(template);
 
-  const encodedStartDate = startDate ? (encodeURIComponent(startDate)) : "";
-  const encodedEndDate = endDate ? (encodeURIComponent(endDate)) : "";
-
   // Showing the last 30 complete days to ensure the dashboard isn't including today which has no data yet
   let encodedSubmissionDate = "30+day+ago+for+30+day";
-  if (encodedStartDate && encodedEndDate) {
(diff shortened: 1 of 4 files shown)


lib/messageUtils.ts
```diff
@@ -58,12 +60,20 @@ export function getDashboard(
   const encodedBranchSlug = branchSlug ? (encodeURIComponent(branchSlug)) : "";
   const dashboardId = getDashboardIdForTemplate(template);
 
+  // Showing the last 30 complete days to ensure the dashboard isn't including today which has no data yet
+  let encodedSubmissionDate = "30+day+ago+for+30+day";
+  if (startDate && endDate) {
+    encodedSubmissionDate = `${startDate}+to+${endDate}`;
+  } else if (startDate) {
+    encodedSubmissionDate = `${startDate}+to+today`;
+  }
```
sarahhjchung, line 69: `branchInfo.nimbusExperiment.endDate` is always null for experiments (I'm assuming because the experiments are still live). In this case, is showing data from `startDate` to today the correct outcome? Or should I show data from `startDate` to the proposed end date, similar to how they're calculated for the date column inside the Skylight tables?

dmose: I'd say use the proposed end date, except when that is in the past, in which case, just use today.

dmose: And by "today", I really mean "the most recent data with data", which I guess is generally yesterday.

lib/messageUtils.ts
```diff
@@ -58,12 +60,20 @@ export function getDashboard(
   const encodedBranchSlug = branchSlug ? (encodeURIComponent(branchSlug)) : "";
   const dashboardId = getDashboardIdForTemplate(template);
 
+  // Showing the last 30 complete days to ensure the dashboard isn't including today which has no data yet
+  let encodedSubmissionDate = "30+day+ago+for+30+day";
+  if (startDate && endDate) {
+    encodedSubmissionDate = `${startDate}+to+${endDate}`;
```
dmose, line 66: One thing to be careful of (and probably manually test to see if the docs are correct, and probably add a comment in the code for future changers of the code):

From https://cloud.google.com/looker/docs/filter-expressions#basic_structure_of_date_and_time_filters

> The initial time value is inclusive but the latter time value is not. So the expression 2018-05-18 12:00:00 to 2018-05-18 14:00:00 will return data with the time "2018-05-18 12:00:00" through "2018-05-18 13:59:59".

For example, I think the above expression would want to effectively be `endDate + 1`, though we'd have to convert to a date type to do the math.

dmose:
In general, this looks good.  I'll go over some of the pieces in more detail after you these changes have been addressed, since I expect some of the details to change.

lib/messageUtils.ts
sarahhjchung: Yeah I noticed this in the dashbaords. I added a logic to +1 to the `proposedDuration` when calculated the `endDate`

dmose: Please do encode all of the URL components, just to be safe (I think an earlier iteration of the patch did this).  The reasoning is that we don't closely track where this comes from, and (for example) the dates came from the network. Although we should be able to trust the network, we don't want to be vulnerable to injection attacks that were started upstream from us.

lib/nimbusRecipe.ts
```diff
@@ -181,10 +181,22 @@ this._rawRecipe.localizations?.[Object.keys(this._rawRecipe.localizations)[0]])
...
+      branchInfo.id,
+      undefined,
+      branchInfo.nimbusExperiment.slug,
+      branch.slug,
+      branchInfo.nimbusExperiment.startDate,
+      getProposedEndDate(
+        branchInfo.nimbusExperiment.startDate,
+        branchInfo.nimbusExperiment.proposedDuration
+          ? branchInfo.nimbusExperiment.proposedDuration + 1
+          : undefined
```
dmose, line 197: To make things more readable, please hoist the `getProposedEndDate` call into its own `const` above this statement so that the reader doesn't have to deal with multiple levels of abstraction as well as the ternary operator as part of understanding the main `getDashboard` invocation.

dmose (approved):
Thanks for the patch; it looks good!  r=dmose with a couple of minor tweaks addressed.

PR #267, opened 2024-06-27
WIP: make inline CTR numbers cover the entire dashboard length

dmose:
This branch is based against the current version of https://github.com/mozilla/skylight/pull/266 and will need to be merged or rebased once that lands.

So far: 

* Looker:
  * Created an element for the about:welcome dashboard that shows CTR and impressions for the dashboard duration
  * Forked the about:welcome dashboard
  * Inserted the CTR/impression element into the forked dashboard
  * Removed the "Other" line
  * Clarified and corrected labels
* This branch
  * Updated to use the new dashboard fork
  * Added code to get the inline CTR info data from the new element
    * This works for dashboards in the release section because they happen to be the 30-day length

Soo...

I think the main things left to do are:
* Make this work with the experiment dates, analogous to those changes in #266.

* handle the infobar dashboard (although since there are no infobar experiments currently, we could potentially push this off to another bug)
** Fork the infobar dashboard and update the code to point to it
** make a similar CTR/impressions element and add it to the forked dashboard
** Update the code to pull its data from the new summary element and handle 

One possibility would be to write the tests in another bug immediately after, just to get this landed.

Because there seems to be a lot of room for error in copy-pasting and then adjusting all the experiment date-handling code in various places, it might be faster to switch to TDD now. Figuring out the right tests to write is not obvious to me at this moment, but maybe after I've slept on it.

### DIFF
--- a/lib/looker.ts
+++ b/lib/looker.ts
@@ -5,15 +5,35 @@ import { getDashboardIdForTemplate } from "./messageUtils";
 export async function getAWDashboardElement0(template: string): Promise<IDashboardElement> {
   const dashboardId = getDashboardIdForTemplate(template);
 
-  // XXX switch this out for the more performant dashboard_element (see 
+  // XXX switch this out for the more performant dashboard_element (see
   // https://mozilla.cloud.looker.com/extensions/marketplace_extension_api_explorer::api-explorer/4.0/methods/Dashboard/dashboard_element
   // for more info).
-  const elements: IDashboardElement[] = await SDK.ok(SDK.dashboard_dashboard_elements(dashboardId));
+
+  let elements: IDashboardElement[];
+  if (template == "infobar") {
+    // XXX infobar hasn't had its own overview dashboard element created yet
+    // so we're still using the old code here
+    elements = await SDK.ok(SDK.dashboard_dashboard_elements(dashboardId));
+  } else {
+    elements = await SDK.ok(
+      // XXX whether search_dashboard_elements is a net win here isn't
+      // clear, but the code is working, so I'm inclined to leave it alone for now.
+      SDK.search_dashboard_elements(
+        {
+          dashboard_id: dashboardId,
+          title: 'CTR and User Profiles Impressed',
+          fields: 'query'
+        }
+      )
+    )
+  }
+
   return elements[0];
 }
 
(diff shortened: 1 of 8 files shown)


lib/looker.ts
```diff
@@ -5,15 +5,35 @@ import { getDashboardIdForTemplate } from "./messageUtils";
 export async function getAWDashboardElement0(template: string): Promise<IDashboardElement> {
   const dashboardId = getDashboardIdForTemplate(template);
 
-  // XXX switch this out for the more performant dashboard_element (see 
+  // XXX switch this out for the more performant dashboard_element (see
```
dmose, line 8: Add a "maybe" to this line?

lib/nimbusRecipeCollection.ts
```diff
@@ -17,14 +18,22 @@ async function updateBranchesCTR(recipe: NimbusRecipe): Promise<BranchInfo[]> {
   return await Promise.all(
     recipe.getBranchInfos().map(
       async (branchInfo: BranchInfo): Promise<BranchInfo> => {
+        const proposedEndDate = getProposedEndDate(
+          branchInfo.nimbusExperiment.startDate,
+          branchInfo.nimbusExperiment.proposedDuration
+            ? branchInfo.nimbusExperiment.proposedDuration + 1
```
dmose, line 24: Need a comment here about why the +1.

lib/looker.ts
```diff
@@ -74,14 +104,16 @@ export async function getCTRPercent(
...
   } else {
+
     queryResult = await runQueryForTemplate(template, {
       "event_counts.message_id": "%" + id + "%",
       "event_counts.normalized_channel": channel,
       "onboarding_v1__experiments.experiment": experiment,
       "onboarding_v1__experiments.branch": branch,
-    });
+    }, startDate, endDate);
+    console.log("queryResult: ", queryResult);
```
dmose, line 116: Let's remove this.

dmose (commented):
This looks good to me.  Please spin out a bug to refactor the date code into something like `getExperimentDashboardDates` with unit tests.  With that done, and the nits I've mentioned address, r=dmose

dmose:
Also, please push your updated tests first.

sarahhjchung (approved):

PR #271, opened 2024-06-28
Bug 1905204 - Refactor date logic for Looker dashboards and add/fix unit tests

sarahhjchung:
[**Bug 1905204**](https://bugzilla.mozilla.org/show_bug.cgi?id=1905204)[Bug 1905204](https://bugzilla.mozilla.org/show_bug.cgi?id=1905204)

Changes made:
- Refactored the experiment dashboard end date logic and added unit tests for the new function `getExperimentDashboardDates`
- Refactored the submission timestamp date logic for Looker dashboard links and added unit tests for the new function `getSubmissionTimestampDateFilter`


lib/nimbusRecipeCollection.ts
```diff
@@ -18,14 +18,9 @@ async function updateBranchesCTR(recipe: NimbusRecipe): Promise<BranchInfo[]> {
   return await Promise.all(
     recipe.getBranchInfos().map(
       async (branchInfo: BranchInfo): Promise<BranchInfo> => {
-        // We are using the proposed end date + 1 as the end date because the end
-        // date is not inclusive in Looker
-        // XXX refactor proposedEndDate into a separate function (see https://bugzilla.mozilla.org/show_bug.cgi?id=1905204)
-        const proposedEndDate = getProposedEndDate(
+        const endDate = getExperimentDashboardDates(
```
dmose, line 21: Do we want to call this `proposedEndDate` for clarity?

lib/experimentUtils.ts
```diff
@@ -106,3 +106,13 @@ export function _substituteLocalizations(values: any, localizations?: any) : obj
 
   return substituted;
 }
+
+/**
+ * @returns the proposed end date used for experiment Looker dashboard links. 
+ *          We are adding 1 to the date because the end date in the Looker
+ *          date filters are not inclusive. 
+ */
```
dmose, line 114: I propose this wants to live in `looker.ts`, perhaps even called `getExperimentLookerDashboardDates` to be clear that there are Looker-specific semantics here. I want to keep the Looker-related stuff isolated to that file in case we ever decide we want to swap out what we used for dashboarding.

lib/messageUtils.ts
```diff
@@ -133,3 +123,19 @@ export function getDashboardIdForTemplate(template: string) {
     return "1806";
   }
 }
+
+/**
+ * @returns the submission timestamp date filter to be used in the Looker dashboard links
+ */
+export function getSubmissionTimestampDateFilter(startDate?: string | null, endDate?: string | null): string {
```
dmose, line 130: This seems like it also wants to live in `looker.ts` for separation-of-concern reasons.

lib/messageUtils.ts
```diff
@@ -133,3 +123,19 @@ export function getDashboardIdForTemplate(template: string) {
...
+  // Showing the last 30 complete days to ensure the dashboard isn't including today which has no data yet
+  let submission_timestamp_date = "30 day ago for 30 day";
+  
+  if (startDate && endDate && (new Date() < new Date(endDate))) {
+    submission_timestamp_date = `${startDate} to ${endDate}`;
+  } else if (startDate) {
+    submission_timestamp_date = `${startDate} to today`;
```
dmose, line 137: Good work separating encoding logic out here.

__tests__/lib/experimentUtils.test.ts
```diff
@@ -124,3 +124,30 @@ describe('_substituteLocalizations', () => {
...
+  it("returns the correct end date when startDate and proposedDuration are defined", () => {
+    const startDate = "2024-06-28";
+    const proposedDuration = 10;
+
+    const result = getExperimentDashboardDates(startDate, proposedDuration);
+
+    expect(result).toEqual("2024-07-09");
+  });
+
+  it("returns the null when startDate is null", () => {
```
dmose, line 138:
```suggestion
  it("returns null when startDate is null", () => {
```

__tests__/lib/experimentUtils.test.ts
```diff
@@ -124,3 +124,30 @@ describe('_substituteLocalizations', () => {
...
+    const proposedDuration = 10;
+
+    const result = getExperimentDashboardDates(null, proposedDuration);
+
+    expect(result).toBeNull();
+  });
+
+  it("returns the null when proposedDuration is undefined", () => {
```
dmose, line 146:
```suggestion
  it("returns null when proposedDuration is undefined", () => {
```

__tests__/lib/messageUtils.test.ts
```diff
@@ -129,9 +129,43 @@ describe('getDashboard', () => {
...
-    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Submission+Timestamp+Date=2024-03-08+to+today&Message+ID=%25${encodeURIComponent(msgId)}%25&Normalized+Channel=&Experiment=&Branch=`
+    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Submission+Timestamp+Date=2024-03-08%20to%20today&Message+ID=%25${encodeURIComponent(msgId)}%25&Normalized+Channel=&Experiment=&Branch=`
 
     const result = getDashboard(template, msgId, undefined, undefined, undefined, startDate, null)
     expect(result).toEqual(expectedLink)
   })
 })
+
+describe("getSubmissionTimestampDateFilter", () => {
+  it("returns the appropriate date filter when startDate and endDate are null", () => {
```
dmose, line 140:
```suggestion
  it("returns the default date filter when startDate and endDate are null", () => {
```

dmose: It would be helpful to be explicit about what behavior is expected for each of the rest of the tests so that if one of the tests breaks, it's immediately clear what's wrong.

dmose (changes requested):
Looking good; no major changes proposed, but since there will be some code motion in addressing the comments, I'd like one more look at it before landing.

lib/experimentUtils.ts
sarahhjchung: Currently, adding this function and also `getSubmissionTimestampDateFilter` into `looker.ts` is causing the `Module not found` errors. So I'm looking into a fix and also an alternative solution.

sarahhjchung: I think the issue is because `looker.ts` has functions that use node modules that should only exist in server components but the functions are used and passed through `columns.tsx` which has `"use client"`. Would it be reasonable to add these functions then to a separate file instead (like `lookerUtils.ts`) to avoid these errors?

sarahhjchung: Yeah, that makes sense

lib/lookerUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+// modules that should only exist in server components. But these functions are 
+// passed through other functions that eventually go through columns.tsx which has
+// "use client". We should look fo a better architectural solution regarding 
+// the situation with these errors.
+
+/**
+ * @returns the proposed end date used for experiment Looker dashboard links. 
+ *          We are adding 1 to the date because the end date in the Looker
+ *          date filters are not inclusive. 
+ */
```
dmose, line 15: Please document the cases where this can return null.

lib/lookerUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+export function getExperimentLookerDashboardDates(startDate : string | null, proposedDuration : number | undefined) {
```
dmose, line 16: Since this doesn't actually return multiple dates, how about renaming this `getLookerDashboardProposedEndDate` for clarity when reading calling sites?

lib/lookerUtils.ts
```diff
@@ -0,0 +1,35 @@
+import { getProposedEndDate } from "./experimentUtils";
+
+// XXX getExperimentLookerDashboardDates and getLookerSubmissionTimestampDateFilter
+// were moved to this file rather than looker.ts due to "Module not found" errors.
+// These errors were likely made because looker.ts has functions that use node
+// modules that should only exist in server components. But these functions are 
+// passed through other functions that eventually go through columns.tsx which has
+// "use client". We should look fo a better architectural solution regarding 
+// the situation with these errors.
```
dmose, line 9: Probably good to say why here.  Something along the lines of doing this both for separation of concerns/code maintainability, but also so that we pre-render more of the markup on the server for performance reasons.

dmose (approved):
Looks good; just a few tweaks requested. Thanks for making this happen; r=dmose with comments addressed.

lib/lookerUtils.ts
sarahhjchung: That makes sense 👍

PR #279, opened 2024-07-02
Bug 1901030 - Enable Prettier

sarahhjchung:
[**Bug 1901030**](https://bugzilla.mozilla.org/show_bug.cgi?id=1901030)

Changes made:
- Added a new step to our CI workflow to check for Prettier warnings
- Installed Prettier and set up npm for it

Note:
- Will wait for review on this set-up before running Prettier on all files


package.json
```diff
@@ -8,7 +8,9 @@
     "start": "next start",
     "lint": "next lint",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "prepare": "husky",
+    "prettier:check:ci": "npx prettier . --check"
```
dmose, line 13:
```suggestion
    "prettier:check:ci": "npx --no prettier . --check"
```
This will prevent npx from trying to do anything updating.

dmose (commented):
This looks good to me. I do want to play around with it some more before giving it final review. I'd suggest adding instructions to set up husky to the beginning of CONTRIBUTING.md, since each develop will want to do that as they get started.

CONTRIBUTING.md
```diff
@@ -7,11 +7,20 @@ practices solidify.
...
 
+## Linting
+
+To correctly run [Prettier](https://prettier.io/docs/en/install) as a pre-commit hook, set up [husky](https://github.com/typicode/husky):
+
+```
+npm i
+npx husky init
+```
+
```
sarahhjchung, line 23: This section is the only part I added to CONTRIBUTING, the rest was the pre-commit linting.

package.json
```diff
@@ -8,7 +8,9 @@
     "start": "next start",
     "lint": "next lint",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "prepare": "husky",
+    "prettier:check:ci": "npx --no prettier . --check"
```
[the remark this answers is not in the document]
sarahhjchung: Good point, I think I just used the full thing because I saw it on a some documentation that I was following but I'll change it to just "prettier"

CONTRIBUTING.md
```diff
@@ -7,7 +7,7 @@ practices solidify.
```
sarahhjchung, line 7: All changes here are from prettier

sarahhjchung:
I've removed husky and lint-staged now. I'll wait to run prettier until the new changes are reviewed

package.json
```diff
@@ -8,7 +8,8 @@
     "start": "next start",
     "lint": "next lint",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "prettier": "npx --no prettier . --check"
```
dmose, line 12: Actually, I realize now that we probably _do_ want this to be"prettier:check", because we want a "prettier" script as well to run it without `--check`, probably something like:

`npm run prettier . --`

The -- tells npm to pass any remaining command-line arguments through to prettier.

I'd suggest we just ditch `npx` entirely.

dmose (commented):
This is looking good.  After you commit the reformatting changes, you'll want to add a file to tell `git blame/annotate` to ignore that commit.  
https://gist.github.com/kateinoigakukun/b0bc920e587851bfffa98b9e279175f2 has details.

dmose:
I suspect you'll want to do the work where you merge from `main`, do the rewrite, and add the ignore file` and then have me do the final review after landing any other in-flight patches that you have, so that you don't have to deal with formatting-related conflicts. Ping me on Slack when you're ready to do that.

dmose:
This looks good.  https://prettier.io/docs/en/install.html suggests adding a .prettierrc that editors and other tools can key off to know that we're using prettier, so once you do that, I think things should be set here.

dmose (approved):
Looks good to me.  I've added a couple of settings to the vscode workspace file so that it should Just Work and run things through prettier on save.

r=dmose

PR #281, opened 2024-07-04
Bug 1901273 - Add info buttons to Metrics column

sarahhjchung:
[**Bug 1901273**](https://bugzilla.mozilla.org/show_bug.cgi?id=1901273)

Changes made:
- Added info buttons next to the Metrics column header to describe CTR values


dmose (changes requested):
This looks good, but somehow it seems to have broken the tooltip for "Messages Released on Firefox", which is weird, since this code hasn't touched that.

I'd propose writing a test for the thing that got broken.  Before doing that, I'd suggest adding [@testing-library/user-event](https://testing-library.com/docs/user-event/intro/), which I've been meaning to do for a while, because it's said to avoid a bunch of unnecessary pain.

Finally, I'd suggest changing the broken tooltip to an InfoPopover, and, if we're lucky, we'll get the fix for free by doing that.

sarahhjchung:
To me, the tooltip next to "Messages Released on Firefox" is still working as it was. I think it just tends to have more of a delay before displaying the tooltip content on hover. But I like the idea of changing it from a tooltip to a popover for this one 👍

dmose (changes requested):
This looks good.  You did a great job following my instructions, which didn't entirely make sense.  Sorry about that!

In particular, we are no longer using `infotooltip.tsx`, so please remove that implementation.

I think you can effectively rename `infotooltip.test.tsx` to `infopopover.test.tx`, more or less search and replace "tooltip" for "popover", and you should have a test that almost but not quite works.  After a fixup or three, I would guess this will be ready to land.

dmose (approved):
Looks good; thanks!  r=dmose

dmose:
Also, please add a CHANGELOG entry; no need for me to re-review.

PR #290, opened 2024-07-11
Bug 1901275 - Improve performance by getting dashboard elements once per dashboard

sarahhjchung:
[**Bug 1901275**](https://bugzilla.mozilla.org/show_bug.cgi?id=1901275)

Changes made:
- Added a new function to get the skeleton dashboard element query per dashboardId

Note:
- Will add tests for `getSkeletonQueryForTemplate` after review of current draft


dmose:
I tested using `rm -fr .next && /bin/time npm run build` and it doesn't make things faster, so I added a bit of logging, and I see at least one reason why.  I forgot to take one thing into account when I proposed the code design: all of the queries are running in parallel, so many of the calls get the uncached skeleton query (until at least one skeleton query per template has returned, which takes a while).

The are various ways to approach this, but I need to prep for a call, so I'll need to think about this some more.

dmose:
I did some poking around here. First, I added some more timed logging to this PR (which I've pushed). As I was playing around with this, I looked in more detail into the pattern of the NextJS worker threads being killed and restarted.  Turns out the default timeout there was 60s, so what was often happening was that the generation would start and get killed multiple times on my Macbook, often not finishing at all, and taking > three minutes of wall clock time.

I tried adding

```diff
+  staticPageGenerationTimeout: 120,
`

This helped a bunch, because it meant that on my machine, anyway, it finished every time. So I'd suggest creating a separate PR to do that and landing it, as it seems like a high-leverage win. Feel free to land rs=dmose (rubber-stamp :-) without review.

If that is enough to improve your dev feedback loop such that you can live with it, I'm somewhat inclined to set this patch aside for the moment, but I'll leave that up to you.

The overall results of my testing looked like this:

```
1: (timeout 60s) Restarted: 1,2,3 (FAIL)
2: (timeout 60s) Restarted: 1,2,3 (FAIL)
3: (timeout 60s) Restarted: 1,2,3 (FAIL)
4 (timeout 120s):
npm run build  28.39s user 3.79s system 39% cpu 1:20.54 total
5: (timeout 120s):
npm run build  28.41s user 3.68s system 37% cpu 1:25.06 total
6: (branch + 120s)
npm run build  28.30s user 3.93s system 40% cpu 1:20.09 total
7: (branch + 120s)
npm run build  28.18s user 3.83s system 41% cpu 1:16.80 total
```

Notably, this patch still didn't seem to have a big effect yet. 

Pre-fetching the (two) template queries and prepopulating the cache with them rather than using memoization will have some effect, but it's unclear to me how much. If that doesn't yield a meaningful win, I think we're going to end up needing to do some more profiling and perhaps looking more closely at when and where our awaits are blocking.

**Comment by dmose** [comment]
BTW, to generate those latest numbers, I was using 

```bash
(rm -fr .next && time npm run build ) |& tee log
```

so that I ended up with a log I could look at.

PR #293, opened 2024-07-12
Bug 1906933 - Improve usability for info and expand/collapse buttons

sarahhjchung:
[**Bug 1906933**](https://bugzilla.mozilla.org/show_bug.cgi?id=1906933)

Changes made:
- Added a Button wrapper to InfoPopover
- Updated styling for InfoPopover and expand/collapse buttons

### DIFF
--- a/components/ui/infopopover.tsx
+++ b/components/ui/infopopover.tsx
@@ -4,6 +4,7 @@ import {
   PopoverTrigger,
 } from "@/components/ui/popover";
 import { Info } from "lucide-react";
+import { Button } from "./button";
 
 type InfoProps = {
   iconSize: number;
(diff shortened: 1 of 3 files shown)


dmose:
I like the idea of making them look and behavior like the "Copy Preview URL" button. What do you think?

sarahhjchung:
I tried playing around with the buttons. I got to something like this to make it look more closely to the preview URL button: 
Screenshot 2024-07-12 at 1 52 33 PM

But I'm not sold on this look as it feels unnatural/clunky to me...

dmose:
I see your point about the clunkiness of having this kind of styling in the small affordances. What I'm mostly looking for is just some sort of visual affordance that doesn't require a mouseover. Any ideas in that direction?

sarahhjchung:
I added a "subtle" background color on the info and expand/collapse buttons for some visual affordance that will darken on hover. I thought it looked less clunky since lacking a border, but still visible.

dmose:
That looks great! 

I noticed that the click target for the info buttons is only the icon, not the entire background area, which feels off; if you could change that, that would be great.

AllegroFox (changes requested):
This is great, I love the circle as a design choice! I agree with Dan about making the info button click area cover the background; looks like we can do this pretty easily by bringing the button element inside PopoverTrigger.

components/ui/infopopover.tsx
```diff
@@ -6,21 +6,15 @@ import {
...
-  iconSize: number;
   content: any;
   iconStyle?: string;
 };
 
-export function InfoPopover({ iconSize, content, iconStyle }: InfoProps) {
+export function InfoPopover({ content, iconStyle }: InfoProps) {
   return (
     <Popover>
       <PopoverTrigger asChild>
```
AllegroFox, line 16: `<Popover>
      <PopoverTrigger asChild>
        <Button variant="outline" size="icon" className={iconStyle}>
          <Info size={iconSize} role="img" aria-label="Info" />
        </Button>
      </PopoverTrigger>
      <PopoverContent className="w-80">
        <div className="grid gap-4">{content}</div>
      </PopoverContent>
    </Popover>`
    
    Seems to work for the clickable area

sarahhjchung: I just saw this after pushing my changes! 😮 I was able to make the background clickable by playing around with the styling and removing the Button wrapper.

AllegroFox (approved):
Thanks! :)

PR #294, opened 2024-07-12
Bug 1907186 - Add timeline component to clarify relationship between tables

sarahhjchung:
[**Bug 1907186**](https://bugzilla.mozilla.org/show_bug.cgi?id=1907186)

Changes made:
- Added a timeline component with stages: Experiment, Rollout, In Tree

### DIFF
--- a/components/ui/timeline.tsx
+++ b/components/ui/timeline.tsx
@@ -1,42 +1,32 @@
+type TimelineStepProps = {
+  isActive: boolean;
+  stepName: string;
+};
+
 type TimelineProps = {
   active: string;
 };
 
+function TimelineStep({ isActive, stepName }: TimelineStepProps) {
+  const activeColorStyle = isActive
+    ? "border-blue-300 text-blue-500"
+    : "border-grey text-slate-500";
+
+  return (
+    <div className={`text-sm border-2 px-2 rounded-full ` + activeColorStyle}>
+      {stepName}
+    </div>
+  );
+}
+
 export function Timeline({ active }: TimelineProps) {
   return (
     <div className="flex flex-row items-center justify-center mt-2">
-      <p
-        className={
-          `text-sm border-2 px-2 rounded-full ` +
-          (active === "experiment"
-            ? "border-blue-300 text-blue-500"
-            : "border-grey text-slate-500")
-        }
-      >
-        Experiment
-      </p>
+      <TimelineStep isActive={active === "experiment"} stepName="Experiment" />
       <hr className="border-1 w-4" />
-      <p
-        className={
-          `text-sm border-2 px-2 rounded-full ` +
-          (active === "rollout"
-            ? "border-blue-300 text-blue-500"
-            : "border-grey text-slate-500")
-        }
-      >
-        Rollout
-      </p>
+      <TimelineStep isActive={active === "rollout"} stepName="Rollout" />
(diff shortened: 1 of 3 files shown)


components/ui/timeline.tsx
```diff
@@ -0,0 +1,42 @@
+type TimelineProps = {
+  active: string;
+};
+
+export function Timeline({ active }: TimelineProps) {
+  return (
+    <div className="flex flex-row items-center justify-center mt-2">
+      <p
```
dmose, line 8: Out of curiosity, why make this a `<p>` rather than a `<div>`, which seems semantically more appropriate?

components/ui/timeline.tsx
```diff
@@ -0,0 +1,42 @@
+type TimelineProps = {
+  active: string;
+};
+
+export function Timeline({ active }: TimelineProps) {
+  return (
+    <div className="flex flex-row items-center justify-center mt-2">
```
dmose, line 7: How about making one circled word a functional component and using that three times for readability & maintainability?

components/ui/timeline.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <p
+        className={
+          `text-sm border-2 px-2 rounded-full ` +
+          (active === "experiment"
+            ? "border-blue-300 text-blue-500"
+            : "border-grey text-slate-500")
```
dmose, line 13: I suggest pulling the ternary expression out into a Javascript variable assignment before the return. This would make the code easier to skim by avoiding having to switch levels of abstraction in the middle of a line.

dmose (changes requested):
This looks great; thanks for making it happen so quickly!  A few comments are inline.

components/ui/timeline.tsx
sarahhjchung: Ah, it just totally blanked my mind that I can add text to `<div>`! I agree that `<div>` seems more appropriate here

components/ui/timeline.tsx
dmose: Looks great!

components/ui/timeline.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function Timeline({ active }: TimelineProps) {
+  return (
+    <div className="flex flex-row items-center justify-center mt-2">
+      <TimelineStep isActive={active === "experiment"} stepName="Experiment" />
+      <hr className="border-1 w-4" />
+      <TimelineStep isActive={active === "rollout"} stepName="Rollout" />
+      <hr className="border-1 w-4" />
+      <TimelineStep isActive={active === "in tree"} stepName="In Tree" />
```
dmose, line 29: How about "In Firefox" instead of "In Tree"?  I think this will be clearer to folks who don't know the detailed mechanics.

components/ui/timeline.tsx
dmose: Great, thanks!

dmose (approved):
This looks great!  If you could add a couple of simple integration-y tests of Timeline to make sure that the active & inactive states are working, that'd be great.  I don't need to re-review after you do.  r=dmose

components/ui/timeline.tsx
sarahhjchung: Yes, I like that better too!

PR #306, opened 2024-07-19
Bug 1903834 - Improve message surface tag readability

sarahhjchung:
[**Bug 1903834**](https://bugzilla.mozilla.org/show_bug.cgi?id=1903834)

Changes made: 
- Added bg colours to the surface tags

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -9,18 +9,18 @@ import { InfoPopover } from "@/components/ui/infopopover";
 
 function getSurfaceTagColor(template: string): string {
   const surfaceTagColors: any = {
-    aboutwelcome: "bg-red-500",
-    defaultaboutwelcome: "bg-orange-500",
-    feature_callout: "bg-yellow-500",
-    infobar: "bg-lime-400",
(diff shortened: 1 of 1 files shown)


dmose:
That looks nice! I find the InfoBar, in particular, hard to read. Please pick a color that is more contrasted to the font color. https://app.contrast-finder.org/?lang=en may be helpful.

AllegroFox:
I agree on the Infobar colour and possibly the Feature Callout one as well (the yellow/gold colour). I think lowering the saturation & brightness on some of the more vivid colours will help with the contrast and readability.

It's worth noting though that _a lot_ of our labels don't pass contrast guidelines for accessibility right now, so while I think fixing these couple of colours for this patch is important, it would also be nice to do a full a11y pass on our UI and tweak things to be in the suggested ranges. (Some of the labels are likely fine because they have other indicators besides contrast, but it's worth taking a look and making some informed choices.) :)

AllegroFox (approved):
This looks good! Thanks for addressing the feedback :)

PR #307, opened 2024-07-19
Bug 1907413 - Link surface tags to experimenter.info docs

sarahhjchung:
[**Bug 1907413**](https://bugzilla.mozilla.org/show_bug.cgi?id=1907413)

Changes made:
- Link surface tags to the appropriate docs with screenshots of the surface
- Fixed surface tags to not wrap
- Added surface tags to WNP messages

Note:
- This PR is branched off from #306 so it will need to be rebased before merging

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -61,7 +61,7 @@ function SurfaceTag(template: string, surface: string) {
         className={
           "text-xs/[180%] " +
           surfaceTagClassName +
-          " visited:text-secondary hover:text-blue-800 hover:bg-opacity-80 cursor-pointer no-underline"
+          " visited:text-secondary hover:text-primary hover:bg-opacity-80 cursor-pointer no-underline"
         }
         href={surfaceDoc}
         target="_blank"
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -7,6 +7,23 @@ import { ChevronsUpDown, ChevronDown, ChevronRight } from "lucide-react";
 import { PrettyDateRange } from "./dates";
 import { InfoPopover } from "@/components/ui/infopopover";
 
+function getSurfaceDocs(template: string): string | undefined {
+  const surfaceTagDocs: any = {
+    feature_callout:
+      "https://experimenter.info/messaging/desktop-messaging-surfaces/#feature-callouts",
+    infobar:
+      "https://experimenter.info/messaging/desktop-messaging-surfaces/#infobar",
+    pb_newtab:
+      "https://experimenter.info/messaging/desktop-messaging-surfaces/#privatebrowsing",
+    spotlight:
+      "https://experimenter.info/messaging/desktop-messaging-surfaces/#multistage-spotlight",
+  };
+
+  if (template in surfaceTagDocs) {
+    return surfaceTagDocs[template];
+  }
+}
+
 function getSurfaceTagColor(template: string): string {
   const surfaceTagColors: any = {
     aboutwelcome: "bg-red-400",
@@ -37,9 +54,27 @@ function SurfaceTag(template: string, surface: string) {
       "px-2 py-1 inline rounded-md " + getSurfaceTagColor(template);
   }
 
-  return (
-    <div className={"text-xs/[180%] " + surfaceTagClassName}>{surface}</div>
-  );
+  const surfaceDoc = getSurfaceDocs(template);
+  if (surfaceDoc) {
+    return (
+      <a
+        className={
+          "text-xs/[180%] " +
+          surfaceTagClassName +
+          " visited:text-secondary hover:text-blue-800 hover:bg-opacity-80 cursor-pointer no-underline"
+        }
+        href={surfaceDoc}
+        target="_blank"
+        rel="noreferrer"
+      >
+        {surface}
+      </a>
+    );
+  } else {
+    return (
+      <div className={"text-xs/[180%] " + surfaceTagClassName}>{surface}</div>
+    );
+  }
 }
 
 function OffsiteLink(href: string, linkText: any) {
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -6,55 +6,17 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
 import { ChevronsUpDown, ChevronDown, ChevronRight } from "lucide-react";
 import { PrettyDateRange } from "./dates";
 import { InfoPopover } from "@/components/ui/infopopover";
-
-function getSurfaceDocs(template: string): string | undefined {
-  const surfaceTagDocs: any = {
-    feature_callout:
-      "https://experimenter.info/messaging/desktop-messaging-surfaces/#feature-callouts",
-    infobar:
-      "https://experimenter.info/messaging/desktop-messaging-surfaces/#infobar",
-    pb_newtab:
-      "https://experimenter.info/messaging/desktop-messaging-surfaces/#privatebrowsing",
-    spotlight:
-      "https://experimenter.info/messaging/desktop-messaging-surfaces/#multistage-spotlight",
-  };
-
-  if (template in surfaceTagDocs) {
-    return surfaceTagDocs[template];
-  }
-}
-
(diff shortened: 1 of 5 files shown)


app/columns.tsx
```diff
@@ -7,6 +7,77 @@ import { ChevronsUpDown, ChevronDown, ChevronRight } from "lucide-react";
 import { PrettyDateRange } from "./dates";
 import { InfoPopover } from "@/components/ui/infopopover";
 
+function getSurfaceDocs(template: string): string | undefined {
```
dmose, line 10: I feel like this wants to be in messageUtils.ts, so that we don't need to go all over the place when adding a surface.  Probably also true for the surfaceTagDocs, now that I think about it.

How about making any array of records indexed by the template identifier and with members including tag, docLink, and color and putting it in that file?

sarahhjchung:
I'm getting rendering issues with the bg-colours when I move the `surfaceTagColors` to `messageUtils.ts` from `columns.tsx`. The bg-colors only seem to show up when the colors are defined inside `columns.tsx` which is why I tried this indexing method. I'm not entirely satisfied though so I'm open to suggestions and feedback on other approaches I could try. I also added fixes here for the surface tag wrapping and adding the WNP surface since those were really small.

lib/messageUtils.ts
```diff
@@ -1,26 +1,70 @@
...
-    multi: "1st of Multiple Messages",
-    pb_newtab: "Private Browsing New Tab",
-    protections_panel: "Protections Dropdown Panel",
-    toast_notification: "Toast Notification",
-    toolbar_badge: "Toolbar Badge",
-    spotlight: "Spotlight Modal Dialog",
-    update_action: "Moments Page",
+export type SurfaceData = {
+  surface: string;
+  tagColorIndex: number;
```
AllegroFox, line 5: If we want to move tagColor here, it'll be 
`tagColor: string;`

lib/messageUtils.ts
```diff
@@ -1,26 +1,70 @@
...
+  docs?: string;
+};
+
```
AllegroFox, line 8: So since we're kind of already manually defining each surface, we can directly define `tagColor` here, and then we just have to make sure we're accessing it as part of SurfaceData in columns.tsx, for ex:

```
aboutwelcome: {
    surface: "About:Welcome Page (1st screen)",
    tagColor: "bg-red-400",
  },
```

lib/messageUtils.ts
```diff
@@ -1,26 +1,70 @@
...
-  if (template in displayNames) {
-    return displayNames[template];
+  if (template in surfaceData) {
+    return surfaceData[template];
   }
 
-  return template;
+  return {
+    surface: template,
+    tagColorIndex: -1,
```
AllegroFox, line 66: we can remove `tagColorIndex` since `tagColor` is in `SurfaceData` now

app/columns.tsx
```diff
@@ -6,40 +6,55 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
...
-  if (template in surfaceTagColors) {
-    return surfaceTagColors[template];
-  }
-
-  return "";
-}
+import { getSurfaceDataForTemplate } from "@/lib/messageUtils";
 
 function SurfaceTag(template: string, surface: string) {
+  const tagColors = [
```
AllegroFox, line 12: instead of defining the tagColors here, we can fetch them out of SurfaceData (along with the docs) like this:

`let { tagColor, docs } = getSurfaceDataForTemplate(template) }`

app/columns.tsx
```diff
@@ -6,40 +6,55 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
...
+    "bg-indigo-400",
+    "bg-purple-400",
+    "bg-pink-400",
+    "bg-rose-400",
+    "bg-fuchsia-300",
+  ];
+
   let surfaceTagClassName = "";
-  if (template !== "none") {
+  const tagColorIndex = getSurfaceDataForTemplate(template).tagColorIndex;
```
AllegroFox, line 29: So now that we're fetching the `tagColor` from SurfaceData, we won't need to compute the `tagColorIndex` here, we can just directly go

`let surfaceTagClassName = "px-2 py-1 inline rounded-md " + tagColor;`

app/columns.tsx
```diff
@@ -6,40 +6,55 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
...
     surfaceTagClassName =
-      "px-2 py-1 inline rounded-md " + getSurfaceTagColor(template);
+      "px-2 py-1 inline rounded-md " + tagColors[tagColorIndex];
   }
 
-  return (
-    <div className={"text-xs/[180%] " + surfaceTagClassName}>{surface}</div>
-  );
+  const surfaceDoc = getSurfaceDataForTemplate(template).docs;
+  if (surfaceDoc) {
```
AllegroFox, line 36: Same as `tagColor` above, now we can directly use `docs` here, like:

```
if (docs) { 
...
```

app/columns.tsx
```diff
@@ -6,40 +6,55 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
...
+    return (
+      <a
+        className={
+          "text-xs/[180%] text-nowrap " +
+          surfaceTagClassName +
+          " visited:text-primary hover:text-secondary hover:bg-opacity-80 cursor-pointer no-underline"
+        }
```
AllegroFox, line 43: here I think this should be `visited:text-primary hover:text-secondary hover:bg-opacity-80 cursor-pointer hover:no-underline` - the `no-underline` is getting parsed as a separate rule for some reason, and taking away the underline all the time. I think we do want it there to indicate the link, and only remove it on hover.

AllegroFox (changes requested):

AllegroFox:
Hey Sarah! I added some suggestions for how to move the tagColors into `messageUtils.ts`, which should help remove some of the computation happening :) let me know if you need better explanations for anything!

app/columns.tsx
sarahhjchung: That's a good point! I'll keep the underline when it's not on hover

sarahhjchung:
Update: moving the `tagColor` to `messageUtils.ts` was still resulting in odd behaviour of the bg-color not showing up. So I reverted those changes back to where I have the `tagColorIndex` for review again.

app/columns.tsx
```diff
@@ -6,40 +6,30 @@ import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
...
 
 function SurfaceTag(template: string, surface: string) {
-  let surfaceTagClassName = "";
-  if (template !== "none") {
-    surfaceTagClassName =
-      "px-2 py-1 inline rounded-md " + getSurfaceTagColor(template);
-  }
+  const { tagColor, docs } = getSurfaceDataForTemplate(template);
+  const surfaceTagClassName =
+    "text-xs/[180%] text-nowrap px-2 py-1 inline rounded-md " + tagColor;
```
AllegroFox, line 14: Minor nit for clarity - now that we know this works, we can tidy it up a bit by putting both the re-used classes into variables, and have another `const` for the anchor tag:
```
const anchorTagClassName = "visited:text-primary hover:text-secondary hover:bg-opacity-80 cursor-pointer hover:no-underline "
```
and then below on line 20 we can just have `anchorTagClassName + surfaceTagClassName`

AllegroFox: Having just played with the preview a bit, I think we should also add `text-primary` to the anchor tag in front of `visited:text-primary` so that we override the default blue on un-visited links...I think the underline and hover state are enough to indicate the link-ness, and the blue fights with the colors & makes for bad contrast in some cases. I'm open to opinions on this though!

AllegroFox (approved):
R+ with the nits addressed! Great work on this :)

app/columns.tsx
sarahhjchung: Yeah, that's a good point. I didn't realize this because I visited all the links already

PR #314, opened 2024-07-24
Bug 1907420 - Show all completed experiments/rollouts on a separate page

sarahhjchung:
[**Bug 1907420**](https://bugzilla.mozilla.org/show_bug.cgi?id=1907420)

Changes made:
- Sorted all experiment/rollouts by date, which are now in yyyy/MM/dd format
- Added surface name for `cfr_doorhanger` features
- Added a new `EXPERIMENTER_API_CALL` for completed experiments to be used in fetchRecipes()
- Added a new page in the route `/complete` to view all the completed experiments/rollouts
- Updated CHANGELOG 

Note:
- Will need to be rebased after #307 is merged due to the surface tag changes


sarahhjchung:
Do we want to add some kind of button to link to the new `/complete` route? I'm not so sure where the best place would be to add this. Some ideas I have are:
- Adding a similar button to the header
- Adding a link inside the Messaging Info menu button
- Having a toggle button in the header (complete/current)
- Adding a link somewhere near the Rollouts and Experiment table title sections

dmose:
Exciting!  First thoughts: 

* having the affordance to find/access it in the menu strikes me as hard to discover/remember.   One half-formed idea that I could imagine (but needs some work) would be to have (Completed) nodes forking off from (Experiment) and (Rollout) in the lifecycle diagram on both pages and have then be visually pressable buttons that would lead to the completed page.  And I guess all of the nodes could be buttons that could link to the appropriate anchors.  Anyway, I think this would need some playing with to see what feels better/write, but their might be something here.

* I'd suggest leaving the rows of both tables collapsed on the completed pages just because there are so many things in each table.

dmose:
Also, I think that (at least) the completed page wants some basic automated testing.

sarahhjchung:
I moved the link to the `/complete` page outside the Messaging Info menu button and into the header, where it's more visible. After some playing around, I think that adding more nodes to the timeline component will make it look too complicated/cluttered, so I don't have any affordance by the table title sections currently. In order to push these changes sooner than later, I think I can just create a bug for more affordances for the future. 

In terms of tests, I've added some more tests to the `MenuButton` component since we already have tests for `MessageTable`.

AllegroFox (approved):
This is a ton of work. Thank you!

sarahhjchung:
Thanks Emily! I just discovered an issue with running build with the Looker turned on so I'm trying to fix that before merging. I may have to ask for a review again after I fix it.

sarahhjchung:
@AllegroFox All the Looker calls were causing the build to break on the new `/completed` page. So for now, I have it so we only provide dashboard links in the Metrics column of the completed rollouts table. I excluded the Metrics column entirely for the completed experiments table because those links were broken for some reason. I'm open to more suggestions on how to approach this and I can also create a spike ticket to look further into the Looker issue for the `/completed` page just to move along with this ticket first.

AllegroFox:
> @AllegroFox All the Looker calls were causing the build to break on the new `/completed` page. So for now, I have it so we only provide dashboard links in the Metrics column of the completed rollouts table. I excluded the Metrics column entirely for the completed experiments table because those links were broken for some reason. I'm open to more suggestions on how to approach this and I can also create a spike ticket to look further into the Looker issue for the `/completed` page just to move along with this ticket first.

Off the top of my head I suspect that the way we're calculating dates for the Looker tables may get broken by using dates that are in the past - I'll take a look and see if it's something obvious like that. In the meantime I think it's OK to do that work in another ticket so that we can land this one.

AllegroFox (approved):

PR #327, opened 2024-08-02
Bug 1909233 - Add search experiment/rollout briefs link to header

sarahhjchung:
[**Bug 1909233**](https://bugzilla.mozilla.org/show_bug.cgi?id=1909233)

Changes made:
- Added a link in the header that leads to the Google Drive with all experiment briefs

Questions:
- Is the header getting too crowded? Not sure if I should move "Message Surfaces" link inside the "Messaging Info" button since surface tags will link to the same doc. 
- Is "Search briefs" informative enough?

### DIFF
--- a/components/ui/menubutton.tsx
+++ b/components/ui/menubutton.tsx
@@ -10,6 +10,7 @@ import {
   NavigationMenuList,
   NavigationMenuTrigger,
   navigationMenuTriggerStyle,
+  navigationMenuItemStyle,
 } from "@/components/ui/navigation-menu";
 import { Menu, Hash, Book, AppWindow, Table, FileSearch } from "lucide-react";
 import { cn } from "@/lib/utils";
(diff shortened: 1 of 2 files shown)


AllegroFox (approved):
This works! LGTM :)

I think as far as the header being crowded - I would be tempted to put both "Messaging Surfaces" and "Technical Docs" inside the Messaging Info dropdown, since they're both documentation links. That way we kind of have "actions" visible across the top, and docs/info in the dropdown. Feel free to do that here, or put it in another PR!

sarahhjchung:
I like that idea! I updated the header to do that. I know previously Dan has given feedback that the items inside the "Messaging Info" button can be harder to find. So I also added a border around the "Messaging Info" button to add some more visibility that it's a menu on hover. But what are your thoughts on that? I'm open to any other ideas too

AllegroFox (approved):
I like it! I think the division of the items makes a lot of sense, and the button border helps indicate that there's more going on with the info item. I think we should go ahead with this and we can ask Dan for feedback on it when he's back!

PR #358, opened 2024-08-26
Bug 1907769 - Microsurvey badging

sarahhjchung:
[**Bug 1897769**](https://bugzilla.mozilla.org/show_bug.cgi?id=1897769)

Changes made:
- Added Microsurvey badges next to messages and experiments/rollouts that have a "survey" sub-string in its id, slug, description, or user facing name
- Made some tiny usability changes that I noticed when demo-ing:
    - Sorted experiments/rollouts in descending order instead of ascending
    - Added padding to the header in the `/complete` page to avoid overflow of tables when scrolling


sarahhjchung:
Note that `groups` aren't being sent through Experimenter API so I wasn't able to look for the micro-survey group. Might be helpful to file a ticket to request for Experimenter API to send the `groups` property to add that check.

lib/nimbusRecipe.ts
```diff
@@ -271,7 +292,8 @@ export class NimbusRecipe implements NimbusRecipeType {
       experimenterLink: `https://experimenter.services.mozilla.com/nimbus/${this._rawRecipe.slug}`,
       userFacingName: this._rawRecipe.userFacingName,
       nimbusExperiment: this._rawRecipe,
-      branches: this.getBranchInfos(),
+      branches: branchInfos,
+      isMicrosurvey: isMicrosurvey,
```
dmose, line 296: How about renaming this to`hasMicrosurvey` here and throughout the patch, since the survey may only be a part of the experiment (like the current review checker experiments).

app/page.tsx
```diff
@@ -74,6 +74,11 @@ async function getASRouterLocalColumnFromJSON(
     channel,
   );
 
+  // Microsurvey check
+  if (fxmsMsgInfo.id.toLowerCase().includes("survey")) {
+    fxmsMsgInfo.isMicrosurvey = true;
+  }
```
dmose, line 80: This feels like it probably wants to be in a `messageHasMicrosurvey` function in messageUtils.ts to keep the implementation details like message parsing heuristics together with message-related functions rather than in page.tsx.  Over time, I think there's a bunch of things that we probably want to move to their own files, probably including the JSON reading stuff.  No rush, however.

lib/nimbusRecipe.ts
```diff
@@ -92,6 +92,18 @@ export class NimbusRecipe implements NimbusRecipeType {
...
+      branchInfo.id.toLowerCase().includes("survey") ||
+      branchInfo.slug.toLowerCase().includes("survey") ||
+      (branchInfo.description &&
+        branchInfo.description.toLowerCase().includes("survey")) ||
+      (branchInfo.userFacingName &&
+        branchInfo.userFacingName.toLowerCase().includes("survey"))
+    ) {
+      branchInfo.isMicrosurvey = true;
+    }
+
```
dmose, line 106: `getBranchInfo` is already too long, let's hoist this into a `_branchInfoHasMicrosurvey()` private method and call it from here.

dmose (changes requested):
Thanks for the patch!

This looks good. I've requested some maintainability-related changes, and it would be good to also add a CHANGELOG entry.

I'd also like to see two tickets be filed as part of landing this:

* getting experimenter JSON API exposing groups (can be filed in mozilla/experimenter issues)
* a bugzilla bug to move us away from having to use heuristics to figure out if messages contain microsurveys

sarahhjchung:
Tickets have been filed here:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1915917 
- https://github.com/mozilla/experimenter/issues/11265

AllegroFox (approved):
Dan covered everything I was going to mention; thanks for the PR!

PR #346, opened 2024-08-19
[WIP] Bug 1906939 - Complete list of live messages

sarahhjchung:
[Bug 1906939](https://bugzilla.mozilla.org/show_bug.cgi?id=1906939)

PR #370, opened 2024-09-03
Bug 1914647 - Using browser tests to collect JSON for live messages 

sarahhjchung:
[**Bug 1914647**](https://bugzilla.mozilla.org/show_bug.cgi?id=1914647)

Changes made:
- Added live message data for releases 123-129
- Added a script to merge all the release JSON data 
- Documented steps on how to run browser tests and commit JSON data into Skylight in CONTRIBUTING.md

Things to consider for future patches: 
- How to handle minor release versions
- How to handle versions 115-123 and versions after 129
- How to handle property collisions (this can a part of [this ticket](https://mozilla-hub.atlassian.net/jira/software/c/projects/OMC/boards/1976/?label=Skylight&selectedIssue=OMC-959))
- I'll also be rebasing this branch against the spike PR https://github.com/mozilla/skylight/pull/346 to compare the live message table between the two PRs


lib/mergeASRouterData.js
```diff
@@ -0,0 +1,33 @@
...
+  let data = fs.readFileSync(
+    `lib/asrouter-local-prod-messages/${release}-release.json`,
+    "utf8",
+  );
+  let json_data = JSON.parse(data);
+
+  // Add any message data with an id that does not already exist in data.json
+  for (let i = 0; i < json_data.length; i++) {
+    if (!json_result.find((x) => x.id === json_data[i].id)) {
+      json_result.push(json_data[i]);
```
dmose, line 25: * the first version read always wins (which is fine if releases stay in decreasing order) - abort script if wrong order, comment near this code that it depends on this ordering
* the preview will always be of the latest released version, even though metrics will cover multiple (possibly all) versions (future epic)
* will need similar algorithm for merge looker data with this data and eventually remote settings data (can be part of spike PR - make note in spike PR)

dmose (approved):
This looks good to me.  r=dmose, once this import script aborts if it sees that the keys aren't in reverse order.

dmose:
Before you land, could you add in 130 as well, since it's the current release version?

I'd also be interested in your thoughts on how we keep this up-to-date.  Might just be a repeating calendar entry or something...

lib/mergeASRouterData.js
sarahhjchung: Instead of aborting script, I just ended up sorting the list in descending order

sarahhjchung:
Yeah, I feel like a repeating calendar entry for every release makes the most sense

lib/mergeASRouterData.js
dmose: Please add a comment that this will need to be checked once if/when we add dot-versions (129.0.1), since I'm not sure how sort's default collation order will handle that...

sarahhjchung: I also had to update the script to start with a basically empty `data.json` (including the default about welcome message that we need to manually add) to account for when newer releases are being added so that they are prioritized first and not being overwritten.

lib/mergeASRouterData.js
```diff
@@ -1,8 +1,640 @@
...
+let availableReleases = [130, 129, 128, 127, 126, 125, 124, 123];
+
+// Sort availableReleases in descending order
+// XXX This will need to be revisited if/once we add dot-versions
+availableReleases.sort(function (a, b) {
+  return b - a;
+});
+console.log(availableReleases);
+
+const manualMessages = [
```
dmose, line 14: Can you clarify why this needs to be manual, and how we go about updating it with each new release?  My suspicion is that even if we need to do some kind of hack to get the data, that hack probably wants to go in the browser test itself, so that we always get the current version.

sarahhjchung: Just talked to @AllegroFox and we figured out how to get the about welcome default message data into the browser test! I'll be pushing my changes soon

sarahhjchung: I've rebased my new changes to the browser test and you can see them here https://phabricator.services.mozilla.com/D201646

CONTRIBUTING.md
```diff
@@ -71,12 +79,65 @@ The Login URI must point at `https://fxms-skylight.netlify.app/api/auth/login`,
...
+
+5. Assuming the test runs successfully, you will find the JSON output in `/tmp/file.json`. Commit the file into Skylight by moving the file to the appropriate location and renaming it in the format `{VERSION_NUM}-release.json`.
+
+   ```bash
+   # In Skylight
+   cd lib/asrouter-local-prod-messages/
+   mv /tmp/file.json ./{VERSION_NUM}-release.json
+   ```
+
+6. Run [`lib/mergeASRouterData.js`](/lib/mergeASRouterData.js) using node to merge all the release data into one file. Make sure you include the latest version number you've just collected data from into the `availableReleases` array inside the script.
```
dmose, line 137: Please split this into two steps, with the first being to edit the script and add the version number, and the second step to run the script.

CONTRIBUTING.md
```diff
@@ -71,12 +79,65 @@ The Login URI must point at `https://fxms-skylight.netlify.app/api/auth/login`,
...
+
+   **❗Important note**: Make sure to update the template for the `MR_WELCOME_DEFAULT` message inside `data.json` to "defaultaboutwelcome". This is currently a manual step we must perform to ensure that the message preview for this unique message is working.
```
dmose, line 139: This line can be deleted now, correct?

dmose (approved):
OK, looks good, once the noted changes have been made.  r=dmose

I've got some stuff in my calendar for early October to see if we can automate ticket creation for this with JIRA.  I'll either do that in October, or fall back to manual calendaring.

PR #378, opened 2024-09-05
Bug 1916839 - Update URL for searching briefs

sarahhjchung:
[Bug 1916839](https://bugzilla.mozilla.org/show_bug.cgi?id=1916839)

Changes made:
- Updated the URL for the "Search Briefs" link in the header to start a search within the experiment briefs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Thursday, September 5th, 2024
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dmose (approved):
Please add a CHANGELOG entry, and this will be ready to land.  I wish I could think of a test to write that wouldn't be super fragile and painful to maintain in our current setup.

r=dmose

PR #380, opened 2024-09-09
Bug 1915881 - Auth0 logging

sarahhjchung:
[Bug 1915881](https://bugzilla.mozilla.org/show_bug.cgi?id=1915881)

Changes made:
- Added a log statement printing the session user's email in both pages (the main page and the completed experiments page)


dmose:
Unfortunately, putting this inside of the page rendering causes all of the dashboards to be rendered at runtime, rather than statically at build-time, meaning a very very long pause on the first load of the dashboard.

It's possible that moving this someone else in the code would help (eg layout.tsx, maybe the middleware).  Let's timebox this, however, and then re-assess:  my current suspicion is that if there's not a simple way to fix it along those lines within an hour or two, that we'd do better to just get access to the Auth0 logs rather than continuing to invest in our own logging of this particular thing. Perhaps you can discuss a bit with @AllegroFox and see what you think.

If one of the above fixes does work in a short period of time, a couple of other thoughts:

* FWIW, I'd ideally like to get one log statement per page load, so that we know about the usage frequency per person, but just getting something to work here is much better than nothing.
* As it stands, this patch seems to fail if IS_AUTH_ENABLED is unset, so we should write a bit of defensive code there.

AllegroFox:
Putting this in a separate comment since it's not really related to my review - I'm thinking it might be time to clean up some of our logging (i.e. of the messages & feature values) since they're getting...big. If we want to be able to see session logs it's getting a bit difficult. 
As an added bonus, removing those logs *significantly* improves the load times of pages!

dmose:
Looks good to me; r=dmose!

Thanks so much for making this happen!

dmose (approved):

PR #384, opened 2024-09-10
Comment out message and feature value related logs

sarahhjchung:
Changes made:
- From https://github.com/mozilla/skylight/pull/380, @AllegroFox discovered that removing these logs will improve the load time for pages locally. I've commented out the logs instead of removing them in case we want to log these with Pino eventually, but that will be part of a separate epic/ticket.


AllegroFox (approved):
Perfect, thank you!

PR #396, opened 2024-09-16
Bug 1914623 - Add toggle option to hide live messages with low impressions

sarahhjchung:
[Bug 1914623](https://bugzilla.mozilla.org/show_bug.cgi?id=1914623)

Changes made:
- Added a checkbox component from shadcn
- Added a `canHideMessages` options for message tables to add a toggle in the Metrics column that hides messages with under 1000 impressions 
    - We don't do anything with messages that don't display any metrics
    - When Looker is not enabled, the toggle won't do anything


app/columns.tsx
```diff
@@ -242,6 +242,15 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
       }
       return <></>;
     },
+    filterFn: (row, columnId, filterValue) => {
+      if (filterValue) {
+        return (
+          (row.original.impressions && row.original.impressions > 1000) ||
```
sarahhjchung, line 248: Any thoughts on 1000 impressions being the threshold? I'm open for any suggestions

dmose: 1000 seems like a reasonable place to start.  Could make it an environment variable so it's easy to play with...

app/message-table.tsx
```diff
@@ -17,19 +20,22 @@ import {
   TableHeader,
   TableRow,
 } from "@/components/ui/table";
-
-import { useState } from "react";
+import { Checkbox } from "@/components/ui/checkbox";
 
 interface MessageTableProps<TData, TValue> {
```
dmose, line 25: This function has gotten large enough with the JSX deeply nested and indented enough that it feels hard to reason about.  I'd like to suggest pulling out the `<TableHead>` JSX invocation into its own functional component (could be in the same file, though). I also wonder if the first `TableBody` invocation wants to be in its own component as well, but I'm more on the fence about that. As long as you're testing the high-level component, I don't this/these internal implementation components would need additional unit testing.

.env.sample
```diff
@@ -38,3 +38,6 @@ LOOKERSDK_CLIENT_SECRET=XXX_REPLACE_WITH_YOUR_CLIENT_SECRET
 LOOKERSDK_VERIFY_SSL=true
 # Timeout in seconds for HTTP requests. Defaults to 2 minutes (120) seconds if not specified.
 LOOKERSDK_TIMEOUT=120
+
+# Threshold for hiding messages with small impression numbers
+IMPRESSIONS_THRESHOLD=1000
```
dmose, line 43: Maybe HIDDEN_MESSAGE_IMPRESSION_THRESHOLD for this and the similar change to variables related to it to make it a bit clearer what this is for when one doesn't have all the context?

app/message-table.tsx
```diff
@@ -57,63 +109,62 @@ export function MessageTable<TData, TValue>({
...
-                      )}
-                    </TableCell>
+    <div>
+      <div className="rounded-md border">
+        <Table>
+          <TableHeader className="sticky top-36">
+            {table.getHeaderGroups().map((headerGroup) => (
+              <TableRow key={headerGroup.id}>
+                {headerGroup.headers.map((header) => {
+                  return MessageTableHead(
```
dmose, line 119: Is there a particular reason not to use the JSX syntax here for consistency/readability?

app/message-table.tsx
```diff
@@ -17,19 +20,65 @@ import {
...
   data: TData[];
   defaultExpanded?: boolean;
+  canHideMessages?: boolean;
+  impressionsThreshold?: string;
+}
+
+function MessageTableHead(
+  header: any,
+  hideMessages: boolean,
+  setHideMessages: any,
```
dmose, line 36: Does this want to be a function type?

app/message-table.tsx
```diff
@@ -17,19 +20,65 @@ import {
...
+  impressionsThreshold?: string,
+) {
+  return (
+    <TableHead className="bg-stone-100 text-slate-400" key={header.id}>
+      {header.isPlaceholder ? null : (
+        <div>
+          {flexRender(header.column.columnDef.header, header.getContext())}
+          {canHideMessages &&
+          header.column.columnDef.header!.toString().includes("Metrics") ? (
+            <div className="flex items-center gap-x-1">
```
dmose, line 47: Overall, even the MessageTabHead function still feels like it has a lot of contextual stuff, and it uses ternary operators (which are hard to read when they're separated in space by a lot of other stuff, but in general, ternary operators are just harder to read than the `if` statements unless their super short).

What do you think about hoisting this `<div>` out into its own functional component as well?

dmose (approved):

app/message-table.tsx
sarahhjchung: Yeah, I think because the `flexRender` logic was there in the original code and part of tanstack tables in general, it might make more sense to hoist the `<div>` part out. The `<div>` part contains most of the checkbox toggle logic anyway! I'll also try to clean up the ternary operation too!

app/message-table.tsx
sarahhjchung: Nope, I can change it to use JSX syntax for better consistency and readability!

app/message-table.tsx
sarahhjchung: I left it as a function type because of the arguments I needed to pass into it

app/columns.tsx
```diff
@@ -242,6 +242,16 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
       }
       return <></>;
     },
+    filterFn: (row, columnId, filterValue) => {
+      if (filterValue) {
+        return (
+          (row.original.impressions &&
+            row.original.impressions >= filterValue) ||
+          !row.original.impressions
```
dmose, line 250: It's worth adding a comment here about the "why" of this clause.

app/message-table.tsx
```diff
@@ -17,19 +20,60 @@ import {
   TableHeader,
   TableRow,
 } from "@/components/ui/table";
+import { Checkbox } from "@/components/ui/checkbox";
 
-import { useState } from "react";
+interface HideMessageCheckboxProps {
+  header: any;
+  hideMessages: boolean;
+  setHideMessages: any;
```
dmose, line 28:
```suggestion
  setHideMessages: (shouldHide : boolean) => any,
```

dmose (approved):
Looks good to me, the code is definitely more readable now.  Thanks!

r=dmose with the added comment and tweaked type signature

PR #414, opened 2024-09-26
Bug 1920756 - Implement default sorting by surface tags for production messages

sarahhjchung:
[Bug 1920756](https://bugzilla.mozilla.org/show_bug.cgi?id=1920756)

Changes made:
- Sorts the live message table by surface tags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## Thursday, Spetember 26th, 2024
+## Thursday, September 26th, 2024
 
 ### Added
 
+- Added default sorting by surface tags for the live message table
 - Exposed dashboard links to completed rollouts and experiments
 
 ## Monday, September 16th, 2024
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -31,13 +31,22 @@ const isLookerEnabled = process.env.IS_LOOKER_ENABLED === "true";
 const hidden_message_impression_threshold =
   process.env.HIDDEN_MESSAGE_IMPRESSION_THRESHOLD;
 
-function compareFn(a: any, b: any) {
+function compareDatesFn(a: any, b: any) {
   if (a._rawRecipe.startDate > b._rawRecipe.startDate) {
     return -1;
   } else if (a._rawRecipe.startDate < b._rawRecipe.startDate) {
     return 1;
   }
   // a must be equal to b
+  return 0;
+}
+
+function compareSurfacesFn(a: any, b: any) {
+  if (a.surface < b.surface) {
+    return -1;
+  } else if (a.surface > b.surface) {
+    return 1;
+  }
   return 0;
 }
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+### Added
+
+- Added default sorting by surface tags for the live message table
+
+## Monday, September 16th, 2024
+
+### Added
+
+- Added a toggle option in the Metrics column to hide live messages with low impressions (ie. messages with less than 1000 impressions)
+
```
dmose, line 12: Thanks for updating this!

app/page.tsx
```diff
@@ -41,6 +41,15 @@ function compareFn(a: any, b: any) {
   return 0;
 }
 
+function compareSurfacesFn(a: any, b: any) {
```
dmose, line 44:
```suggestion
function compareSurfacesFn(a: FxMSMessageInfo, b: FxMSMessageInfo): number {
```

As part of our ongoing cleanup of page.tsx, how about moving this to `messageUtil.ts`?

app/page.tsx
```diff
@@ -31,7 +31,7 @@ const isLookerEnabled = process.env.IS_LOOKER_ENABLED === "true";
 const hidden_message_impression_threshold =
   process.env.HIDDEN_MESSAGE_IMPRESSION_THRESHOLD;
 
-function compareFn(a: any, b: any) {
+function compareDatesFn(a: any, b: any) {
```
dmose, line 34:
```suggestion
function compareDatesFn(a: string, b: string): number {
```

dmose (changes requested):
This looks nice; thanks for making it happen!  Got a few nits, and it'd be good to add an automated test or two that covers the added code.  If it's straightforward, an integration test(s) would be ideal, since the code inside the function doesn't itself seem particularly fragile.

app/page.tsx
```diff
@@ -31,12 +32,15 @@ const isLookerEnabled = process.env.IS_LOOKER_ENABLED === "true";
 const hidden_message_impression_threshold =
   process.env.HIDDEN_MESSAGE_IMPRESSION_THRESHOLD;
 
-function compareFn(a: any, b: any) {
-  if (a._rawRecipe.startDate > b._rawRecipe.startDate) {
-    return -1;
-  } else if (a._rawRecipe.startDate < b._rawRecipe.startDate) {
-    return 1;
+function compareDatesFn(a: NimbusRecipe, b: NimbusRecipe): number {
```
dmose, line 35: Please add a JSDoc comment here to make it explicit to the reader that if one or both of the recipes is missing a start date, they will be ordered identically when using this compare function. Describing why this is desirable behavior seems worthwhile too, even if it's merely an arbitrary choice, since this will help future changes consider this.

dmose (approved):
Thanks for the changes, and sorry for the delay. This looks good, r=dmose with comment change addressed.  Note that there is a `CHANGELOG.md` conflict with `main` that will need to be hand merged before landing.

PR #415, opened 2024-09-26
Bug 1921080 - Expose dashboard links to completed messages 

sarahhjchung:
[Bug 1921080](https://bugzilla.mozilla.org/show_bug.cgi?id=1921080)

Changes made:
- Exposed dashboard links to completed rollouts and experiments
- Updated the dashboard date logic
    - The new logic is now to pass in an `isCompleted` parameter to the date calculation functions, and if the message is completed, the dashboard will show data from the startDate to the endDate. Otherwise with the old logic, completed messages would always show dashboards ranging from the startDate to today.

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -7,6 +7,7 @@ import { ChevronsUpDown, ChevronDown, ChevronRight } from "lucide-react";
 import { PrettyDateRange } from "./dates";
 import { InfoPopover } from "@/components/ui/infopopover";
 import { getSurfaceDataForTemplate } from "@/lib/messageUtils";
+import { HIDE_DASHBOARD_EXPERIMENTS } from "@/lib/experimentUtils";
 
 function SurfaceTag(template: string, surface: string) {
   const { tagColor, docs } = getSurfaceDataForTemplate(template);
@@ -426,14 +427,8 @@ export const experimentColumns: ColumnDef<RecipeOrBranchInfo>[] = [
       </div>
     ),
     cell: (props: any) => {
-      // XXX these dashboards are currently (incorrectly) empty.
-      // Until we fix the upcase bug, we'll hide them
-      const hideDashboardExperiments = [
-        "recommend-media-addons-feature-existing-users",
-        "recommend-media-addons-feature-callout",
-      ];
       if (
-        hideDashboardExperiments.includes(
+        HIDE_DASHBOARD_EXPERIMENTS.includes(
           props.row.original?.nimbusExperiment?.slug,
         )
       ) {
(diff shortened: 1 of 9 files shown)


app/columns.tsx
```diff
@@ -607,6 +607,30 @@ export const completedExperimentColumns: ColumnDef<RecipeOrBranchInfo>[] = [
       );
     },
   },
+  {
+    accessorKey: "metrics",
+    header: () => "Metrics",
+    cell: (props: any) => {
```
dmose, line 613: How about hoisting this into `experimentUtils.ts` so that if we want to add or remove experiments from the list, we don't inadvertently miss a place that we need to edit...

lib/lookerUtils.ts
```diff
@@ -30,11 +30,14 @@ export function getExperimentLookerDashboardDate(
 export function getLookerSubmissionTimestampDateFilter(
```
dmose, line 30: If you could add `@param` blocks for each parameter and explain how they affect the expected outcomes, this would allow future code readers to able to tell if a specific code behavior is a bug and more easily understand how to call it.

lib/messageUtils.ts
```diff
@@ -105,14 +105,15 @@ export function getDashboard(
   branchSlug?: string,
   startDate?: string | null,
   endDate?: string | null,
+  isCompleted?: boolean,
```
dmose, line 108: If you could add a comment explaining that this could be useful for messages that used to be in remote settings or old versions of Firefox, that would be nice, because it took me a bit to figure this out, and I wouldn't want the code to be inadvertently removed because of lack of context.

lib/nimbusRecipe.ts
```diff
@@ -240,14 +243,21 @@ export class NimbusRecipe implements NimbusRecipeType {
       branchInfo.nimbusExperiment.startDate,
       branchInfo.nimbusExperiment.proposedDuration,
     );
+    let formattedEndDate;
+    if (branchInfo.nimbusExperiment.endDate) {
+      const endDate = new Date(branchInfo.nimbusExperiment.endDate);
+      endDate.setUTCDate(endDate.getUTCDate() + 1);
+      formattedEndDate = endDate.toISOString().slice(0, 10);
+    }
```
dmose, line 251: If you moved the added clause into a formatDate() function in `experimentUtils.ts`, then this stanza of code would be shorter and more readable/maintainable, because it wouldn't be intermingling different levels of abstraction.

dmose (approved):
This looks good to me; thanks so much for the quick turnaround!  The changes I'm requesting seem like they should be fairly minor, so no need to get another review from me unless there's something you want specific feedback on.  r=dmose

PR #417, opened 2024-09-27
Bug 1917036 - Allow filtering by message surface

sarahhjchung:
[Bug 1917036](https://bugzilla.mozilla.org/show_bug.cgi?id=1917036)

Changes made:
- Added a search bar in the Surface column to allow filtering by surfaces


AllegroFox (commented):
This is great! The filtering as you type is a nice feature.

I'd like to wait for Dan's feedback also; I'm wondering if the size of the search bar should be increased a bit for readability, but I'm not sure what the balance would be with making the table heading too large.

Also, is there a way to prevent the table rows from dynamically resizing themselves? It's noticeable when we expand/collapse branch info on the lower tables but it's not too bad because the size difference isn't huge. However, with the filtering function, the table can change size A LOT depending on the length of the remaining message IDs, and with it constantly changing as you type in the search box it gets pretty jarring. If it's an easy fix, it would be nice to do before landing this.

sarahhjchung:
Those are great points! I'll try playing around with different sizes for the search bar and also see if I can make the columns a fixed width. I'm not sure how easy the latter is so I'll probably have to time box that.

AllegroFox:
So I've played with this a bit and it looks like the only way to stop the filter from changing the table column size is to make the table fixed, and I'm not 100% sure that's something we want to do (we might be able to do a fixed width for the preview button & metrics columns, but I doubt we want that for message/experiment name & branch info columns since the content length can vary so much.) I'll defer to Dan's feedback there :)

app/message-table.tsx
```diff
@@ -132,6 +133,23 @@ export function MessageTable<TData, TValue>({
...
+                                (header.column.getFilterValue() ?? "") as string
+                              }
+                              onChange={(e) =>
+                                header.column.setFilterValue(e.target.value)
+                              }
+                              placeholder={`Search...`}
+                              className="w-full border border-slate-400 font-light p-1 text-2xs rounded"
+                            />
+                          </div>
+                        )}
```
dmose, line 152: For maintainability & readability, how about hoisting this into its own functional component in this file.

dmose (changes requested):
This looks good (and sorry for the delay)! I've added one request, and I'd like to see a few basic automated tests as well.

dmose:
Although I agree that it would be nice to avoid the jumpiness of the autocompletion, I don't feel like it's a big deal. We can always file another ticket if I'm wrong...

sarahhjchung:
I'm getting a weird console error, despite the tests all passing:
``` 
Warning: The current testing environment is not configured to support act(...)
```

Since this is only affecting the way I wrote the tests, I just pushed my changes anyway while I continue to look into this console error and how to rid of it.

dmose (approved):
Thanks for the updated patch.  Here's some comments/questions.  As far as the `act` stuff, after reviewing https://medium.com/@AbbasPlusPlus/react-testing-library-understanding-act-and-when-to-use-it-301bd06fd1bc, I played around a little bit and got something that works by pulling the render down into the test itself:

```typescript
      it("filters messages by surface without any case sensitivity", async () => {
        const user = userEvent.setup();

        const { findByRole } = render(
          <MessageTable
            columns={fxmsMessageColumns}
            data={[fxmsMsgInfo1, fxmsMsgInfo2, fxmsMsgInfo3]}
          />,
        );
        const surfaceFilterTextBox = await findByRole("textbox");
        await user.type(surfaceFilterTextBox, "feature");

        const featureCalloutMsg = screen.queryByText("MESSAGE_1");
        const infoBarMsg = screen.queryByText("MESSAGE_2");
        const defaultAboutWelcomeMsg = screen.queryByText("MESSAGE_3");
        expect(featureCalloutMsg).toBeInTheDocument();
        expect(infoBarMsg).not.toBeInTheDocument();
        expect(defaultAboutWelcomeMsg).not.toBeInTheDocument();
      });
```

I kinda wonder if it wouldn't be more correct to use findByText for at least some `queryByText` as well, and pull those up into the execute section.

dmose (changes requested):
Whoops, I would like to see this again before landing.

app/message-table.tsx
```diff
@@ -38,21 +46,21 @@ interface MessageTableProps<TData, TValue> {
...
   return (
     <div className="flex items-center gap-x-1">
       <Checkbox
-        className="border-slate-500"
+        className="border-slate-400"
         id="hide"
         onCheckedChange={() => {
           if (!hideMessages) {
-            header.column.setFilterValue(parseInt(impressionsThreshold!));
+            column.setFilterValue(parseInt(impressionsThreshold!));
```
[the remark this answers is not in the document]
sarahhjchung: I think it was because the type for environment variables was automatically `string | undefined` but I knew it was going to be defined in `.env`. But I agree the type assumptions should be more consistent, so I added a XXX comment there

app/message-table.tsx
```diff
@@ -177,3 +180,41 @@ export function MessageTable<TData, TValue>({
...
+  const [hideMessages, setHideMessages] = useState(false);
+
+  if (filterVariant === "text") {
+    return (
+      <div>
+        <input
+          type="text"
+          value={(column.getFilterValue() ?? "") as string}
+          onChange={(e) => column.setFilterValue(e.target.value)}
+          placeholder={`Search...`}
```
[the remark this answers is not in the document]
sarahhjchung: That's a good point!

__tests__/app/message-table.test.tsx
```diff
@@ -534,22 +536,16 @@ describe("MessageTable", () => {
...
+        messages = [fxmsMsgInfo1, fxmsMsgInfo2, fxmsMsgInfo3];
       });
 
       it("filters messages by surface without any case sensitivity", async () => {
         const user = userEvent.setup();
-        const surfaceFilterTextBox = screen.getByRole("textbox");
-        await act(async () => {
-          await user.type(surfaceFilterTextBox, "feature");
-        });
+        render(<MessageTable columns={fxmsMessageColumns} data={messages} />);
```
dmose, line 544: Please add a blank line above this to separate the setup and execute steps.

__tests__/app/message-table.test.tsx
```diff
@@ -560,14 +556,13 @@ describe("MessageTable", () => {
 
       it("filters messages by surface for any substring", async () => {
         const user = userEvent.setup();
-        const surfaceFilterTextBox = screen.getByRole("textbox");
-        await act(async () => {
-          await user.type(surfaceFilterTextBox, "1st");
-        });
+        render(<MessageTable columns={fxmsMessageColumns} data={messages} />);
```
dmose, line 559: Please add a blank line above this to separate the setup and execute steps.

dmose (approved):
Looks great; r=dmose with a couple of nits.  Thanks so much!

PR #446, opened 2024-10-18
Add support for editing messages with Limelight, where possible

AllegroFox:


dmose:
When I try to run this branch locally, I see this: 

grafik

AllegroFox:
@dmose Is it possible you've got some malformed JSON on your local? I'm only able to trigger that if I do something like remove the screens array from a message, or put a "spotlight" template on something that isn't a spotlight fr. ex. 
I can wrap this function in a try/catch or something so that messages with wonky JSON don't generate edit links :)

dmose:
> @dmose Is it possible you've got some malformed JSON on your local? I'm only able to trigger that if I do something like remove the screens array from a message, or put a "spotlight" template on something that isn't a spotlight fr. ex. I can wrap this function in a try/catch or something so that messages with wonky JSON don't generate edit links :)

@AllegroFox I did some debugging, and it turns out that this data is coming from Looker, despite the function names: https://github.com/mozilla/skylight/blob/main/app/page.tsx#L150-L152 (though, to be fair, think link is to `main`, and I've rebased my copy locally against current `main`).  That said, I just checked a non-rebased version, and still saw the problem.

What happened, I think, is that the functions were written in a world before we got production data from Looker, and when the refactoring introduced that, the function names didn't change.

At the very least we're going to want to rename some functions here so stacks are less confusing.

CC @sarahhjchung

PR #452, opened 2024-10-22
Bug 1889195 - Fix lower-case message id dashboard links

sarahhjchung:
I found out that Looker filters can include OR statements so I tried updating logic so the message ID filters could catch the original ID, lower cased ID, and upper cased ID in all their filters for the dashboard links and CTR metrics. This would ensure that we don't always upper case all IDs and miss any message IDs that are lower case in the telemetry, without having to update anything in the dashboard LookMLs. I have this draft PR up for any initial feedback before I go in to clean up the tests.

### DIFF
--- a/__tests__/lib/messageUtils.test.ts
+++ b/__tests__/lib/messageUtils.test.ts
@@ -84,7 +84,7 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
     expect(params.get("Messaging System Ping Type")).toBe(template);
     expect(params.get("Submission Date")).toBe(submissionDate);
     expect(params.get("Messaging System Message Id")).toBe(msgId);
@@ -103,8 +103,8 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
-    expect(params.has("Submission Timestamp Date", submissionDate)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
+    expect(params.get("Submission Timestamp Date")).toBe(submissionDate);
     expect(params.get("Message ID")).toBe(`%${msgId}%`);
     expect(params.get("Normalized Channel")).toBe("");
     expect(params.get("Experiment")).toBe("");
@@ -130,7 +130,7 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
     expect(params.get("Submission Timestamp Date")).toBe(submissionDate);
     expect(params.get("Message ID")).toBe(`%${msgId}%`);
     expect(params.get("Normalized Channel")).toBe("");
@@ -158,7 +158,7 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
     expect(params.get("Submission Timestamp Date")).toBe(submissionDate);
     expect(params.get("Message ID")).toBe(`%${msgId}%`);
     expect(params.get("Normalized Channel")).toBe("");
@@ -186,7 +186,7 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
     expect(params.get("Submission Timestamp Date")).toBe(submissionDate);
     expect(params.get("Message ID")).toBe(`%${msgId}%`);
     expect(params.get("Normalized Channel")).toBe("");
@@ -213,7 +213,7 @@ describe("getDashboard", () => {
     const url = new URL(result);
     const params = url.searchParams;
 
-    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(url.pathname.endsWith(dashboardId)).toBe(true);
     expect(params.get("Submission Timestamp Date")).toBe(submissionDate);
     expect(params.get("Message ID")).toBe(`%${msgId}%`);
     expect(params.get("Normalized Channel")).toBe("");
--- a/__tests__/lib/messageUtils.test.ts
+++ b/__tests__/lib/messageUtils.test.ts
@@ -80,22 +80,17 @@ describe("getDashboard", () => {
       channel,
       experiment,
       branchSlug,
-    );
-    const resultUrl = new URL(result!);
-    const resultParams = new URLSearchParams(resultUrl.search);
+    ) as string;
+    const url = new URL(result);
+    const params = url.searchParams;
 
-    expect(resultUrl.pathname.includes(dashboardId)).toBe(true);
-    expect(resultParams.has("Messaging System Ping Type", template)).toBe(true);
-    expect(resultParams.has("Submission Date", submissionDate)).toBe(true);
-    expect(
-      resultParams.has(
-        "Messaging System Message Id",
-        `${msgId},` + `${msgId.toLowerCase()},` + `${msgId.toUpperCase()}`,
-      ),
-    ).toBe(true);
-    expect(resultParams.has("Normalized Channel", channel)).toBe(true);
-    expect(resultParams.has("Experiment", experiment)).toBe(true);
-    expect(resultParams.has("Experiment Branch", branchSlug)).toBe(true);
+    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(params.get("Messaging System Ping Type")).toBe(template);
+    expect(params.get("Submission Date")).toBe(submissionDate);
+    expect(params.get("Messaging System Message Id")).toBe(msgId);
+    expect(params.get("Normalized Channel")).toBe(channel);
+    expect(params.get("Experiment")).toBe(experiment);
+    expect(params.get("Experiment Branch")).toBe(branchSlug);
   });
 
   it("returns a correct featureCallout dashboard link", () => {
@@ -104,25 +99,16 @@ describe("getDashboard", () => {
     const dashboardId = getDashboardIdForTemplate(template);
     const submissionDate = "30 day ago for 30 day";
 
-    const result = getDashboard(template, msgId);
-    const resultUrl = new URL(result!);
-    const resultParams = new URLSearchParams(resultUrl.search);
+    const result = getDashboard(template, msgId) as string;
+    const url = new URL(result);
+    const params = url.searchParams;
 
-    expect(resultUrl.pathname.includes(dashboardId)).toBe(true);
-    expect(resultParams.has("Submission Timestamp Date", submissionDate)).toBe(
-      true,
-    );
-    expect(
-      resultParams.has(
-        "Message ID",
-        `%${msgId}%,` +
-          `%${msgId.toLowerCase()}%,` +
-          `%${msgId.toUpperCase()}%`,
-      ),
-    ).toBe(true);
-    expect(resultParams.has("Normalized Channel", "")).toBe(true);
-    expect(resultParams.has("Experiment", "")).toBe(true);
-    expect(resultParams.has("Branch", "")).toBe(true);
+    expect(url.pathname.includes(dashboardId)).toBe(true);
+    expect(params.has("Submission Timestamp Date", submissionDate)).toBe(true);
+    expect(params.get("Message ID")).toBe(`%${msgId}%`);
+    expect(params.get("Normalized Channel")).toBe("");
+    expect(params.get("Experiment")).toBe("");
+    expect(params.get("Branch")).toBe("");
   });
 
   // XXX should this be "about:welcome" to be consistent with featureIds?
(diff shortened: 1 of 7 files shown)


lib/looker.ts
```diff
@@ -47,6 +47,7 @@ export async function runLookQuery(lookId: string): Promise<string> {
 export async function runQueryForTemplate(
   template: string,
   filters: any,
+  filter_expression?: any,
```
dmose, line 50: Please add JSDoc specifying what this function does.  Be sure to describe how filters and filter_expression interact.

__tests__/lib/messageUtils.test.ts
```diff
@@ -81,7 +81,7 @@ describe("getDashboard", () => {
       branchSlug,
     );
 
-    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Messaging+System+Ping+Type=${encodeURIComponent(template)}&Submission+Date=30%20day%20ago%20for%2030%20day&Messaging+System+Message+Id=${encodeURIComponent(msgId)}&Normalized+Channel=${encodeURIComponent(channel)}&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=&Experiment=${encodeURIComponent(experiment)}&Experiment+Branch=${encodeURIComponent(branchSlug)}`;
+    const expectedLink = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Messaging+System+Ping+Type=${encodeURIComponent(template)}&Submission+Date=30%20day%20ago%20for%2030%20day&Messaging+System+Message+Id=${encodeURIComponent(msgId)}%2C+${encodeURIComponent(msgId).toLowerCase()}%2C+${encodeURIComponent(msgId).toUpperCase()}&Normalized+Channel=${encodeURIComponent(channel)}&Normalized+OS=&Client+Info+App+Display+Version=&Normalized+Country+Code=&Experiment=${encodeURIComponent(experiment)}&Experiment+Branch=${encodeURIComponent(branchSlug)}`;
```
dmose, line 84: One of the things that I've been struggling with is that it feels like we've hit a tipping point where it's now hard to have confidence that, especially for tests, but also to some degree the code under test, it's hard to have confidence that:

a) the current state is actually correct
b) changes to the current state aren't breaking something

I think that a big part of what's going on here is that we're doing all the encoding and testing of the encoding by hand, at the same time as we're also trying to test that they are creating URLs that do what we want to do here.

Which is another way of saying that I think some separation of concerns would help meaningfully.  More in upcoming comments...

dmose: Would love hear thoughts from you and @AllegroFox

lib/messageUtils.ts
```diff
@@ -131,12 +131,20 @@ export function getDashboard(
   );
   const dashboardId = getDashboardIdForTemplate(template);
 
+  let baseUrl = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}`;
+
   if (_isAboutWelcomeTemplate(template)) {
-    return `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Submission+Timestamp+Date=${encodedSubmissionDate}&Message+ID=%25${encodedMsgId?.toUpperCase()}%25&Normalized+Channel=${encodedChannel}&Experiment=${encodedExperiment}&Branch=${encodedBranchSlug}`;
+    return new URL(
+      `?Submission+Timestamp+Date=${encodedSubmissionDate}&Message+ID=%25${encodedMsgId}%25%2C+%25${encodedMsgId.toLowerCase()}%25%2C+%25${encodedMsgId.toUpperCase()}%25&Normalized+Channel=${encodedChannel}&Experiment=${encodedExperiment}&Branch=${encodedBranchSlug}`,
```
dmose, line 138: I think you did the right thing here based on what I communicated, and this does help validate stuff. That said, I didn't communicate what I had in mind at all clearly.

My thinking is to use not only `URL` but also `URLSearchParams` so that we don't do hand-encoding but focus entirely on getting the right parameters.

Here's a diff that shows a rough approximation of what I have in mind:

```diff
diff --git a/__tests__/lib/messageUtils.test.ts b/__tests__/lib/messageUtils.test.ts
index e5179b9..08a7abb 100644
--- a/__tests__/lib/messageUtils.test.ts
+++ b/__tests__/lib/messageUtils.test.ts
@@ -67,7 +67,7 @@ describe("toBinary", () => {
 describe("getDashboard", () => {
   it("returns a correct infobar dashboard link w/exp & branch", () => {
     const template = "infobar";
-    const msgId = "12`3"; // weird chars to test URI encoding
+    const msgId = "ab`c"; // weird chars to test URI encoding
     const channel = "release";
     const experiment = "experiment:test";
     const branchSlug = "treatment:a";
@@ -87,7 +87,7 @@ describe("getDashboard", () => {
 
   it("returns a correct featureCallout dashboard link", () => {
     const template = "feature_callout";
-    const msgId = "1:23"; // weird chars to test URI encoding
+    const msgId = "a:bc"; // weird chars to test URI encoding
     const dashboardId = getDashboardIdForTemplate(template);
 
     const expectedLink = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Submission+Timestamp+Date=30%20day%20ago%20for%2030%20day&Message+ID=%25${encodeURIComponent(msgId)}%25%2C+%25${encodeURIComponent(msgId).toLowerCase()}%25%2C+%25${encodeURIComponent(msgId).toUpperCase()}%25&Normalized+Channel=&Experiment=&Branch=`;
@@ -99,7 +99,7 @@ describe("getDashboard", () => {
   // XXX should this be "about:welcome" to be consistent with featureIds?
   it("returns a correct aboutwelcome dashboard link w/exp & branch", () => {
     const template = "aboutwelcome";
-    const msgId = "1:23"; // weird chars to test URI encoding
+    const msgId = "a:bc"; // weird chars to test URI encoding
     const experiment = "experiment:test";
     const branchSlug = "treatment:a";
     const dashboardId = getDashboardIdForTemplate(template);
@@ -118,7 +118,7 @@ describe("getDashboard", () => {
 
   it("returns a correct dashboard link with defined start and end dates where the end date is in the future", () => {
     const template = "feature_callout";
-    const msgId = "1:23"; // weird chars to test URI encoding
+    const msgId = "a:bc"; // weird chars to test URI encoding
     const startDate = "2024-03-08";
     const endDate = "2025-06-28";
     const dashboardId = getDashboardIdForTemplate(template);
@@ -139,7 +139,7 @@ describe("getDashboard", () => {
 
   it("returns a correct dashboard link with defined start and end dates where the end date is in the past", () => {
     const template = "feature_callout";
-    const msgId = "1:23"; // weird chars to test URI encoding
+    const msgId = "a:bc"; // weird chars to test URI encoding
     const startDate = "2024-03-08";
     const endDate = "2024-05-28";
     const dashboardId = getDashboardIdForTemplate(template);
@@ -160,7 +160,7 @@ describe("getDashboard", () => {
 
   it("returns a correct dashboard link with a defined start date", () => {
     const template = "feature_callout";
-    const msgId = "1:23"; // weird chars to test URI encoding
+    const msgId = "a:bc"; // weird chars to test URI encoding
     const startDate = "2024-03-08";
     const dashboardId = getDashboardIdForTemplate(template);
 
diff --git a/lib/messageUtils.ts b/lib/messageUtils.ts
index 529c54f..e072211 100644
--- a/lib/messageUtils.ts
+++ b/lib/messageUtils.ts
@@ -119,25 +119,42 @@ export function getDashboard(
   endDate?: string | null,
   isCompleted?: boolean,
 ): string | undefined {
+
   const encodedMsgId = encodeURIComponent(msgId);
+  // XXX comment here about what we're trying to do here and why
+  // each of the three clauses here is required (assuming all three are)
+  const multiCaseMessageIds =
+    `%${msgId}%,` + `%${msgId.toLowerCase()}%,` + `%${msgId.toUpperCase()}%`;
+
   const encodedTemplate = encodeURIComponent(template);
   const encodedChannel = channel ? encodeURIComponent(channel) : "";
   const encodedExperiment = experiment ? encodeURIComponent(experiment) : "";
   const encodedBranchSlug = branchSlug ? encodeURIComponent(branchSlug) : "";
   // The isCompleted value can be useful for messages that used to be in remote
   // settings or old versions of Firefox.
-  const encodedSubmissionDate = encodeURIComponent(
-    getLookerSubmissionTimestampDateFilter(startDate, endDate, isCompleted),
+  const submissionDate = getLookerSubmissionTimestampDateFilter(
+    startDate,
+    endDate,
+    isCompleted,
   );
+  const encodedSubmissionDate = encodeURIComponent(submissionDate);
   const dashboardId = getDashboardIdForTemplate(template);
 
   let baseUrl = `https://mozilla.cloud.looker.com/dashboards/${dashboardId}`;
 
   if (_isAboutWelcomeTemplate(template)) {
-    return new URL(
-      `?Submission+Timestamp+Date=${encodedSubmissionDate}&Message+ID=%25${encodedMsgId}%25%2C+%25${encodedMsgId.toLowerCase()}%25%2C+%25${encodedMsgId.toUpperCase()}%25&Normalized+Channel=${encodedChannel}&Experiment=${encodedExperiment}&Branch=${encodedBranchSlug}`,
-      baseUrl,
-    ).toString();
+    const paramObj = {
+      "Submission Timestamp Date": submissionDate,
+      "Message ID": multiCaseMessageIds,
+      "Normalized Channel": channel ? channel : "",
+      Experiment: experiment ? encodeURIComponent(experiment) : "",
+      Branch: branchSlug ? branchSlug : "",
+    };
+    const params = new URLSearchParams(Object.entries(paramObj));
+
+    let n = new URL(baseUrl);
+    n.search = params.toString();
+    return n.toString();
   }
 
   if (template === "infobar") {
 ```
 
The changes to the tests here were just to make the output more readable so I could convince myself that I wasn't breaking things terribly.

Note that the tests are still failing, but I _think_ that the failures now are all valid variants and still work.  They also make the URLs easier to read which is a nice bonus. 

My current thinking is that we probably want the tests to also avoid dealing with the encoding stuff, but are likely to be reasonable if we switch them over to URL and its accessor functions/

Would love to hear thoughts from you and @AllegroFox on this as well.

**Review by dmose — CHANGES_REQUESTED** [review]

File: lib/messageUtils.ts (hunk 3 above)
Comment (sarahhjchung) [reply lib/messageUtils.ts:138]: Ahh, I see. I agree using `URLSearchParams` makes a lot of sense and also makes it a lot more readable. And yes, after taking a look at the tests after using the `URLSearchParams`, I think that the failures is just due to the encoding being slightly different, but still valid. I'll try playing around with the tests so that the encoding should be consistent.

File: lib/asrouter-local-prod-messages/data.json [hunk 4]
```diff
@@ -774,7 +774,7 @@
   },
   {
     "weight": 100,
-    "id": "INFOBAR_LAUNCH_ON_LOGIN",
+    "id": "infobar_launch_on_login",
```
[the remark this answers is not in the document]
sarahhjchung: Yes, sorry I forgot to change this back 😅

__tests__/lib/messageUtils.test.ts
sarahhjchung: Yeah, this makes sense. I tried updating the tests in `messageUtils.test.ts` so we don't have to write the encoding and the full URLs by hand. Instead, I'm checking the URLSearchParams to check the values for certain parameters which will also do the encoding for me.

sarahhjchung: I'm just not entirely sure how to update the tests in `nimbusRecipe.test.ts` similarly because the tests aren't specifically testing the dashboard links but are failing because they're included in the `branchInfo` objects.

dmose: How about something like this:

```diff
diff --git a/__tests__/lib/nimbusRecipe.test.ts b/__tests__/lib/nimbusRecipe.test.ts
index aded435..bdc359a 100644
--- a/__tests__/lib/nimbusRecipe.test.ts
+++ b/__tests__/lib/nimbusRecipe.test.ts
@@ -1,7 +1,9 @@
 import { NimbusRecipe } from "@/lib/nimbusRecipe";
 import { ExperimentFakes } from "@/__tests__/ExperimentFakes.mjs";
 import { BranchInfo } from "@/app/columns.jsx";
-import { getDashboardIdForTemplate } from "@/lib/messageUtils";
+import { formatDate } from "@/lib/experimentUtils";
+import { getExperimentLookerDashboardDate } from "@/lib/lookerUtils";
+import { getDashboard, getDashboardIdForTemplate } from "@/lib/messageUtils";
 
 //XXX We're passing this custom object for about:welcome, since ExperimentFakes
 // doesn't quite give us what we want in that case. We probably want to tweak
@@ -136,14 +138,35 @@ describe("NimbusRecipe", () => {
       const branch = AW_RECIPE.branches[1];
 
       const branchInfo = nimbusRecipe.getBranchInfo(branch);
+      const proposedEndDate = getExperimentLookerDashboardDate(
+        branchInfo.nimbusExperiment.startDate,
+        branchInfo.nimbusExperiment.proposedDuration,
+      );
       const dashboardId = getDashboardIdForTemplate("aboutwelcome");
+      const formattedEndDate = formatDate(
+        branchInfo.nimbusExperiment.endDate as string,
+        1,
+      );
+
+      const dashboardLink = getDashboard(
+        branch.template as string, // XXX this seems wrong
+        branchInfo.id,
+        undefined,
+        branchInfo.nimbusExperiment.slug,
+        branch.slug,
+        branchInfo.nimbusExperiment.startDate,
+        branchInfo.nimbusExperiment.endDate
+          ? formattedEndDate
+          : proposedEndDate,
+        false, // default for isCompleted in constructor
+      );
 
       // XXX getBranchInfo is actually going to return a previewLink, which
       // makes this test kind of brittle. We could refactor this to no longer
       // use deepEqual and check for the existence of object properties instead.
       expect(branchInfo).toEqual({
         product: "Desktop",
-        ctrDashboardLink: `https://mozilla.cloud.looker.com/dashboards/${dashboardId}?Submission+Timestamp+Date=30%20day%20ago%20for%2030%20day&Message+ID=%25feature_value_id%3Atreatment-a%25%2C+%25${"feature_value_id%3Atreatment-a".toLowerCase()}%25%2C+%25${"feature_value_id%3Atreatment-a".toUpperCase()}%25&Normalized+Channel=&Experiment=aboutwelcome-test-recipe&Branch=treatment-a`,
+        ctrDashboardLink: dashboardLink,
         id: "feature_value_id:treatment-a",
         isBranch: true,
         nimbusExperiment: AW_RECIPE,
```

dmose (changes requested):
So, an interesting development. The primary motivation for making these case-sensitivity changes was because of the two experiments in `HIDE_DASHBOARD_EXPERIMENTS` whose dashboards we were hiding because they had been broken by case-sensitivity issues. 

Just to verify that this made the changes we hoped, I checked out `main`, commented out those two experiments so that `HIDE_DASHBOARD_EXPERIMENTS` was an empty list, and interestingly enough, all 6 branch dashboards worked (I went back to the Completed page to find them). Meaning that some other change that happened in the meantime (either by us, or Data Science, or Looker, or...) fixed the problem we were seeing before.

 I think we also had hoped this might fix some of the things in `hideDashboardMessages`, but I just checked and it does not fixed those (though I started to figure out something there which I will put in the appropriate bug).
 
That said, I think the refactoring we did here makes that code substantially less fragile and easier to understand, so I think we should keep most of this patch.

In addition to emptying (but leaving) the `HIDE_DASHBOARD_EXPERIMENTS`, I think we probably want to get rid of the filter expression, as well as the three different versions of the message id. 

But all the URLSearchParam refactoring seems like a big win.

lib/experimentUtils.ts
```diff
@@ -42,10 +42,7 @@ export const MESSAGING_EXPERIMENTS_DEFAULT_FEATURES: string[] = [
  *
  * XXX Until we fix the upcase bug, we'll be hiding these dashboards.
```
dmose, line 43: How about changing this comment to just say that if we have experiment dashboards with serious problems, we can hide those dashboards by adding the experiment sluges to this array?

dmose (approved):
Looks great; just a few minor nits to tweak, r=dmose for landing.

PR #498, opened 2024-12-09
Move correct version dependabot.yaml to the right place, no bug

dmose:
This is an attempt to make the automatic dependency management group things correctly and work in a reasonable way. The reason this wasn't working before was because I had put the dependabot.yaml file I had worked on in the wrong places. This commit moves it to the right place, and blows away the default file.


sarahhjchung (approved):
Looks good, thanks!

PR #507, opened 2024-12-11
Revert "Bump @types/node from 22.10.1 to 22.10.2"

dmose:
This reverts commit b76028cf385e093b8da7158bf0f0ab3672ad1de2.  That change triggered a crash in the typescript compiler during CI that won't be fixed until TypeScript 5.7.3.  This should fix the "red" status of `main` in CI.


dmose:
Excitingly, reverting the most recent change didn't fix the bug, even though that is when ti was first seen. 

I wonder if Github's CodeQL checker upgraded their TypeScript version to 5.7, which happened to be coincidental in time, and that's what triggered this.

I don't want to disable the CodeQL checker, but....

dmose:
Removing the review request, since it doesn't really make sense to merge this (ie, it won't fix the problem).

PR #524, opened 2024-12-18
Bug 1891121 - Add experiment brief links to experiment and rollout tables

sarahhjchung:
[Bug 1891121](https://bugzilla.mozilla.org/show_bug.cgi?id=1891121)

Changes made:
- `RecipeInfo` now includes an `experimentBriefLink` property with the url to experiment link Google docs
- A tooltip button of the `FileText` icon from lucide-icons is now displayed next to the experiment/rollout name if an experiment brief link exists
- The package `@mozilla/nimbus-schemas` has been imported and replaces the `NimbusExperiment` type used from `@mozilla/nimbus-shared`

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -3,11 +3,22 @@ import { types } from "@mozilla/nimbus-shared";
 import { ColumnDef, Row } from "@tanstack/react-table";
 import { NimbusRecipe } from "@/lib/nimbusRecipe";
 import { PreviewLinkButton } from "@/components/ui/previewlinkbutton";
-import { ChevronsUpDown, ChevronDown, ChevronRight } from "lucide-react";
+import {
+  ChevronsUpDown,
+  ChevronDown,
+  ChevronRight,
+  FileText,
+} from "lucide-react";
 import { PrettyDateRange } from "./dates";
 import { InfoPopover } from "@/components/ui/infopopover";
 import { getSurfaceDataForTemplate } from "@/lib/messageUtils";
 import { HIDE_DASHBOARD_EXPERIMENTS } from "@/lib/experimentUtils";
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from "@/components/ui/tooltip";
 
 function SurfaceTag(template: string, surface: string) {
   const { tagColor, docs } = getSurfaceDataForTemplate(template);
@@ -98,6 +109,7 @@ export type RecipeInfo = {
   isBranch?: boolean;
   branches: BranchInfo[]; // XXX rename this to branchInfos to avoid confusion with the branches property inside NimbusExperiment
   hasMicrosurvey?: boolean;
+  experimentBriefLink?: string;
 };
 
 export type BranchInfo = {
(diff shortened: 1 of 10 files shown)


sarahhjchung:
`DocumentationLinks` were exposed in the experimenter v7 api and I am able to get the data from it. However, I'm getting build failures due to `documentationLinks` not being defined in the `node_modules/@mozilla/nimbus-shared/types/experiments.ts` file. Is this file something that needs to be updated by us or something we need to request the nimbus team for?

dmose:
We need to request that the experimenter team do it (or put in a PR ourselves and ask for their review).

app/columns.tsx
```diff
@@ -370,32 +404,36 @@ export const experimentColumns: ColumnDef<RecipeOrBranchInfo>[] = [
...
+                  aria-hidden="true"
+                  style={{
+                    marginInline: "0.1rem 0",
+                    paddingBlock: "0 0.1rem",
+                    overflow: "visible",
+                  }}
+                >
+                  <path d="m1.71278 0h.57093c.31531 0 .57092.255837.57092.571429 0 .315591-.25561.571431-.57092.571431h-.57093c-.31531 0-.57093.25583-.57093.57143v4.57142c0 .3156.25562.57143.57093.57143h4.56741c.31531 0 .57093-.25583.57093-.57143v-.57142c0-.3156.25561-.57143.57092-.57143.31532 0 .57093.25583.57093.57143v.57142c0 .94678-.76684 1.71429-1.71278 1.71429h-4.56741c-.945943 0-1.71278-.76751-1.71278-1.71429v-4.57142c0-.946778.766837-1.71429 1.71278-1.71429zm5.71629 0c.23083.0002686.43879.13963.52697.353143.02881.069172.04375.143342.04396.218286v2.857141c0 .31559-.25561.57143-.57093.57143-.31531 0-.57092-.25584-.57092-.57143v-1.47771l-1.88006 1.88171c-.14335.14855-.35562.20812-.55523.15583-.19962-.0523-.35551-.20832-.40775-.40811-.05225-.19979.00727-.41225.15569-.55572l1.88006-1.88171h-1.47642c-.31531 0-.57093-.25584-.57093-.571431 0-.315592.25562-.571429.57093-.571429z"></path>
```
dmose, line 431: Can you add an XXX comment both here and later in the file that we should switch to the Lucide Icon for an external link instead of maintaining our own? (I don't think we need to fix this now).

package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@auth0/nextjs-auth0": "^3.6.0",
     "@looker/sdk-node": "25.4.0",
+    "@mozilla/nimbus-schemas": "^3001.0.0",
     "@mozilla/nimbus-shared": "^2.5.2",
```
dmose, line 23: From reading https://github.com/mozilla/nimbus-shared, it sounds like we don't need `nimbus-shared` any more.

lib/nimbusRecipe.ts
```diff
@@ -14,7 +13,10 @@ import {
 } from "../lib/experimentUtils.ts";
 import { getExperimentLookerDashboardDate } from "./lookerUtils.ts";
 
-type NimbusExperiment = types.experiments.NimbusExperiment;
+const nimbusExperimentV7Schema = require("@mozilla/nimbus-schemas/schemas/NimbusExperimentV7.schema.json");
+type NimbusExperiment = typeof nimbusExperimentV7Schema.properties;
+type DocumentationLink =
+  typeof nimbusExperimentV7Schema.properties.documentationLinks;
```
dmose, line 19: I think it would be helpful to put JSDoc comments here explaining why you're doing the type aliasing.

dmose (changes requested):
This is looking good!  A couple of UX things that I think we want to tweak though:

image

1. In all of them, I think the experiment slug wants to be indented to start in the same place as the human-readable experiment name, as it's visually a bit confusing as-is.  That said, that would line the experiment info up with the branch info, so maybe we'd want to indent the branches more too. Play around a bit and see what looks reasonable.
2. In the Mozilla VPN one, if it's easy to prevent wrapping (eg with `&nbsp;`) a line-break just before the symbol, I think the visual presentation would be smoother.

sarahhjchung:
I played around with the spacing and indenting and concluded on this. 

Screenshot 2025-03-24 at 4 39 39 PM

I've made the user facing name experiment/rollout (ie. the bold text) to have the `overflow-visible` property to prevent weird wrapping with the external link icon because I couldn't figure out exactly how to work `&nbsp;` to my advantage. Let me know if the UI tweaks look good and I'm always welcome to more feedback!

lib/messageUtils.ts
```diff
@@ -40,6 +40,10 @@ export function getSurfaceDataForTemplate(template: string): SurfaceData {
       surface: "Menu Messages",
       tagColor: "bg-pink-300",
     },
+    menu_message: {
```
dmose, line 43: There are two different names for the same surface?

dmose (approved):
This looks good to me.  One small question to answer and/or address; I don't need to see this again.  Thanks for making this happen as well as your patience! r=dmose

PR #552, opened 2025-01-16
Fix the Firefox pill in the Timeline component

sarahhjchung:
**Description**

The `firefox` id was removed from a previous patch which caused the Firefox pill in the Timeline component to be broken. 

**Changes made:**

- Added a `firefox` id to the live message table heading and adjusted padding


dmose:
Looks good, but please add a simple test so that this doesn't regress again.

I don't need to look at the test before you merge, unless you specifically want input.

r=dmose

dmose (approved):

PR #555, opened 2025-01-21
Fix for long message IDs that break the UI

sarahhjchung:
**Changes made:**
- Truncates (and adds an ellipses) message IDs in the live message table to be max 50 characters
- Added a simple test to check that truncating occurs for long message IDs

### DIFF
--- a/app/columns.tsx
+++ b/app/columns.tsx
@@ -197,7 +197,7 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
             {/* Quick fix for long message IDs that can break the UI */}
             {props.row.original.id.length <= 50
               ? props.row.original.id
-              : props.row.original.id.substring(0, 50) + "..."}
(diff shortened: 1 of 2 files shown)


app/columns.tsx
```diff
@@ -194,7 +194,10 @@ export const fxmsMessageColumns: ColumnDef<FxMSMessageInfo>[] = [
       return (
         <>
           <div className="font-mono text-xs inline">
-            {props.row.original.id}
+            {/* Quick fix for long message IDs that can break the UI */}
+            {props.row.original.id.length <= 50
+              ? props.row.original.id
+              : props.row.original.id.substring(0, 50) + "..."}
```
dmose, line 200: Nit: please use the Unicode ellipsis character (…) and test for that as well.

dmose (approved):
Nice work, and thanks for the quick turnaround complete with test.  One nit to tweak, r=dmose once you've done so.

PR #577, opened 2025-02-07
Switch NodeJS versions from 20 (prod)/21 (canary) to 22/23

dmose:
Netlify is soon going to default to NodeJS 22.  Let's make sure things are going well there...


sarahhjchung (approved):
Thanks Dan!

PR #585, opened 2025-02-10
Update @auth0/nextjs-auth0 to 3.6.0

dmose:
This fixes a security issue caused by depending on a vulnerable version of the cookie package.


dmose:
Tried this on my own test auth0 account, and it seemed to work nicely.

PR #657, opened 2025-04-10
Update Looker dashboard submission timestamp logic

sarahhjchung:
Changes made:
- Updates `getLookerSubmissionTimestampDateFilter` so that any endDate that is in the future gets set to today's date to avoid issues with the dashboard showing no data for future dates
- Also included a small fix to the external link icon for CTRs to be properly inline


__tests__/lib/messageUtils.test.ts
```diff
@@ -142,9 +142,9 @@ describe("getDashboard", () => {
     const template = "feature_callout";
```
dmose, line 142: Nit: how about clarifying that the expected result here is today?

dmose (approved):
Looks good; please update the changelog too.  r=dmose, with one nit

PR #664, opened 2025-04-16
Changes to the feature ID list, getAndroidBranchInfo, and getAndroidDashboardLink

sarahhjchung:
Changes made: 
- This PR isolates the changes made to the feature ID list, `getAndroidBranchInfo`, and `getAndroidDashboardLink` from https://github.com/mozilla/skylight/pull/653
- This PR also includes some unit tests I've added to test `getAndroidBranchInfo`, and `getAndroidDashboardLink`


sarahhjchung (commented):
Adding some tests for `Nimbus.getBranchInfo`, specifically for `getAndroidBranchInfo` on top of the comments left here

__tests__/lib/nimbusRecipe.test.ts
```diff
@@ -109,124 +155,193 @@ describe("NimbusRecipe", () => {
...
-        undefined,
-        branchInfo.nimbusExperiment.slug,
-        branch.slug,
-        branchInfo.nimbusExperiment.startDate,
-        branchInfo.nimbusExperiment.endDate
-          ? formattedEndDate
-          : proposedEndDate,
-        false, // default for isCompleted in constructor
-      );
+    describe("getAndroidBranchInfo", () => {
```
sarahhjchung, line 281: The commits look really complicated but for this file I just simply added the two tests inside this `describe("getAndroidBranchInfo", () => { ... })`

lib/experimentUtils.ts
```diff
@@ -31,6 +33,13 @@ export const MESSAGING_EXPERIMENTS_DEFAULT_FEATURES: string[] = [
   "spotlight",
   "testFeature",
   "whatsNewPage",
+
+  // Android features
+  "cfr",
```
AllegroFox, line 38: Do we have a way to differentiate between the Desktop cfr feature and the Android one? If they're falling in the same bucket we may run into issues

sarahhjchung: Yeah, I think since we check for the platform by `appName` before deciding to call `getDesktopBranchInfo` or `getAndroidBranchInfo` so I think that would cover it

PR #665, opened 2025-04-17
Update Android Looker dashboard ID

sarahhjchung:
Changes made:
- Added a new tile to the android Looker dashboard https://mozilla.cloud.looker.com/dashboards/2303?Normalized%20Channel=release&Submission%20Date=30%20day&Experiment%20Slug=rootca-info-card-hcr1-fenix&Value%20Branch=&Sample%20ID=%3C%3D10&Value=info%5E_card%5E_rootCA%5E_HCR1%25
- Updated the looker dashboard ID for android so dashboard links point to the updated dashboard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Updated
 
-- Updated the Looker dashboards for Android messaging experiments with a new tile to show total users profile impressed and the primary rate CTR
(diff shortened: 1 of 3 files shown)


dmose (commented):
While I know I was the one guilty of spreading the "pivot_index on column" pattern to the top tile here, I'd prefer not to see it spread further than it already has, now that we know better. Would you please fix both tiles to use `pivot_where`?

sarahhjchung:
Great point! I've updated both tiles to use `pivot_where`

dmose (approved):
Probably also worth including that this fixed that bug in the CHANGELOG.  r=dmose

